PR #8, opened 2020-04-29
Move nginx configuration to an example file

pusewicz:
This PR moves the example nginx configuration to a file for easier installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,14 @@ When klipper is running, open the url `http://<printer-ip>:7125/printer/info` in
 ```
 your API is working!
 
+### Install & Configure Mainsail
+Now you can download the current mainsail static data
+```bash
+mkdir ~/mainsail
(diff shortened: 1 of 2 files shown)


meteyou:
good idea, but i don't like the nginx-config in the release zip. the release files are generated and i would have to copy them back in manually every time (very error-prone every time). can we store them somewhere else in the repo and download them from there?

pusewicz:
One would not have to copy it over and over again, right?

It's definitely less error prone than having to manually copy & paste it from the README file.

I imagine there would also be some install script later on that would take care of all the manual stuff.

meteyou:
there are already some "installers" for mainsail. i have to add them to the readme.md. this should solve the problem with the nginx config.

- https://github.com/ArmyAg08/mainsail-installer
- https://github.com/BlackStump/mainsail-install

PR #38, opened 2020-08-15
Initial development of Power Control Panel

Birkemosen:
Hello meteyou

Thanks for all your work on this nice interface.

I have created a Power Control Panel, based on the latest moonraker api, utilizing the "Power" api.

The panel is only shown when there are defined power devices in moonraker: (printer/power/devices)
To get the status of each device this endpoint is used: (printer/power/status)
Power requests are handled at these endpoints: (printer/power/on ; printer/power/off -  param : { "moonraker device id" } ) 

Hope it is useful.

PowerControlPanel


meteyou:
please do not use a reformat code function of your ide!

Birkemosen:
Sorry for commiting autoformatted. That should be fixed now.

meteyou:
at the moment, i'm working on the new api [https://github.com/Arksine/moonraker/blob/dev-protocol-update/docs/api_changes.md](https://github.com/Arksine/moonraker/blob/dev-protocol-update/docs/api_changes.md).

at first i would finish this work, and then check the PR again.

meteyou:
@Birkemosen now you can update your plugin to the new version.

Birkemosen:
Thanks, I have updated to the new version

meteyou:
can you hide the panel, if there are no power devices exists?
is it possible to change the state with the response and not with the click on the button?

Birkemosen:
Valid points, it should be fixed.

meteyou:
yesterday was hopefully the last API change. can you please adapt the plugin again?

Birkemosen:
I have updated the plugin to match the latest api changes.

skuridin:
@Birkemosen is this happening because the moonraker needs to be updated or missed check?

image

Birkemosen:
As far as i have tested right now it happens because, when you dont have a [power} section in your moonraker config, moonraker doesn't expose the power api -> leading to the "Method not found" error
In the **getPowerDevices** function i actually check for that problem (data.error.code == -32601), which evidently is a little to late since the **onmessage** function is activated before.

The **onmessage** function in _wsClient.js_ handles that error by logging it to console, as far as i can see.

PR #42, opened 2020-08-16
Update installation.md

asovetov:
Not all users use raspbian. Not all users understand that there is a path in the config, and nginx does not start with the wrong path. Not every user is familiar with Linux =)


meteyou:
my installation guide is only for raspian / raspberryPi os. if you use a different distro, then unfortunately the users themselves have to make it work. i can't write a guide for each distro, because then nobody would be able to follow it.

PR #58, opened 2020-09-06
Use encodeURI for paths

kageurufu:
Moonraker expects file paths to use un-escaped file paths,
 such as gcodes/some%20file.gcode, but encodeURIComponent
 will always escape the slash. Using encodeURI here will not.


meteyou:
thx for the input + PR!

PR #64, opened 2020-09-11
feature: allows flipping the webcam horizontally and vertically

cadriel:
- Adds settings to toggle a flip on the X or Y axis of the webcam image.
- Updates the webcam view to apply the above two transforms based on the settings.
- Updates default config to represent both toggles off.
- Moves the webcam URL to the same location as other webcam configuration.
- Added a base `.gitignore` to avoid accidental commits to things like `node_modules`.

@meteyou I wasn't sure what your intent was with the webcam URL, so I moved it to be consistent. Happy to move this back if you feel that's appropriate tho.

Also - I've left the settings in place to allow rotation of the webcam feed, but have hidden the controls for now. Allowing rotation on an image that is not square causes issues, as rotation happens after page layout. I researched a bunch of ways to reposition layout based on the transform - but nothing really appealed to me. IMO - it might be better to not allow rotation at all unless we want to implement a more complex solution to this. I tend to think that if a user needs a 90 or 270 degree rotation, it should be done further upstream.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,12 @@
+node_modules
+.DS_Store
+design
+*.log
+packages/test
+dist
+temp
+.vuerc
+.version
+.versions
+.changelog
+package-lock.json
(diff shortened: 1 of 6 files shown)


meteyou:
thx for your work!

PR #65, opened 2020-09-11
Share the sort function in getters between functions

yanfali:
- factor out sort into a named function and share it among common
   callers
 - ~~add a suggested .editorconfig to avoid changing indents and line endings~~
 - ~~add a suggested .pretterrc which honors original formatting as much
   as possible~~

~~Apologies for formatting changes, these are done automatically by the tooling. Let me know if you want them gone. I generally don't format code by hand anymore and just let prettier take care of it on save.~~

### DIFF
--- a/src/store/getters.js
+++ b/src/store/getters.js
@@ -1,4 +1,12 @@
+function caseInsensitiveNameSort(a, b) {
+    let nameA = a.name.toUpperCase();
+    let nameB = b.name.toUpperCase();
 
+    if (nameA < nameB) return -1;
+    if (nameA > nameB) return 1;
+
+    return 0;
+}
 
 export default {
     heaters: state => {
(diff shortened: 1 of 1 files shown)


meteyou:
please remove the .editorconfig, .pretterrc and undo the formatting. thx for the shared sort function

yanfali:
@meteyou I will do as you ask.

I would just like to mention that this makes contributing patches much more involved. Editorconfig is a editor agnostic standard that helps other people honor your indenting and line ending standards, so if you could introduce one for your project I would be most appreciative. Prettier also helps other people to contribute by confirming to your formatting standards automatically; for example double quotes instead of single quotes.

Many thanks.

src/store/getters.js
```diff
@@ -126,11 +110,11 @@ export default {
     },
 
     fan: state => {
-        return Object.entries(state.printer).indexOf('fan') ? state.printer.fan : false;
```
yanfali, line 129: let me know if you want me to revert these. They just caught my eye when I started reverting. @meteyou

meteyou:
it makes sense to use something like this in the future, but i don't have enough time and want to deal with it. the double quotes are ok. i will merge this PR now.

PR #72, opened 2020-09-15
Support mainsail through an SSL proxy

kageurufu:
I like to run all my web services proxied through a single
 internet-accessible server, secured through SSL and authenticated
 against a single proxy server (nginx, in my case)

I found that mainsail hardcoded ws:// and many http:// addresses, so
 I've updated them to detect ws:// vs ws:///, and just use the protocol
 agnostic // prefix, which then sends the request over the same protocol
 the page is loaded under, either http or https.

Now i've got a nice SSL secured but publicly accessible URL for remote
 printer monitoring

image


meteyou:
looks nice! thx!

PR #75, opened 2020-09-16
Update macros.md

tumbleweedlabs:
Per meteyou, this G91 is erroneous and can be omitted.


meteyou:
i updated all macros and i removed it with this update. thx for your report!

PR #80, opened 2020-09-22
Mainsail logo for mobile site pinning

translucentfocus:
**Before:**  The number 1 is because it pulls it from the IP Address of the site on my LAN
image

**After:**
image


meteyou:
We only need 1 size? Is this for every device? Is PNG the best picture format for this or is a vector picture like SVG better?

translucentfocus:
> 
> 
> We only need 1 size? Is this for every device? Is PNG the best picture format for this or is a vector picture like SVG better?

Great question!  Actually, no, there are different favicons for different devices.  The 180x180 PNG works for iOS and Android when pinning the site to your home screen, but the 32x32 favicon.ico is used by browsers for the tab icon.  Here's some more info [here](https://bitsofco.de/all-about-favicons-and-touch-icons/).

The SVG can be used for the logo on the web page itself, but a PNG is what you'll want to use for the browser favicons.

I chatted with lixxbox and found some issues with the SVG file in your source files (the edges aren't smooth for some reason). Once you have the new SVG file on github, I plan to use the SVG source file to redo both the 32x32 favicon.ico & the 180x180 PNG as it can be a smaller file size and less blurry.

Did that make sense?

PR #83, opened 2020-09-26
Enhancement: Add enabling Gzip for mainsail nginx config

translucentfocus:
Old Pull Request: https://github.com/meteyou/mainsail/pull/78

Revised to place config in `/etc/nginx/sites-available/mainsail`

Also modified `gzip_comp_level` to a value of `4` instead of the default `6` to lower CPU usage... it _may_ now work on the PI Zero without killing the CPU, but I don't have one to test with.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -148,6 +148,17 @@ server {
     listen 80 default_server;
     listen [::]:80 default_server;
 
+    gzip on;
+
+    gzip_vary on;
+    gzip_proxied any;
+    gzip_proxied expired no-cache no-store private auth;
+    gzip_comp_level 4;
+    gzip_buffers 16 8k;
+    gzip_http_version 1.1;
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -148,6 +148,17 @@ server {
...
 
+    gzip on;
+
+    gzip_vary on;
+    gzip_proxied any;
+    gzip_proxied expired no-cache no-store private auth;
+    gzip_comp_level 4;
+    gzip_buffers 16 8k;
+    gzip_http_version 1.1;
+    gzip_types text/plain text/css text/xml text/javascript application/x-javascript application/json application/xml $
```
meteyou, line 159: the $ character at the end, normally indicates a truncated string in nano and in this config no $ at the end should be necessary. please check again if this...

translucentfocus: Ah! Sorry!  Thank you for pointing that out... completely missed it... commits updated!

PR #109, opened 2020-10-25
Filter console messages before adding them to internally

randellhodges:
Add the ability to apply one or more regular expressions to messages to filter out those we do not want to see. It would be nice to have some way to tell klipper/moonraker not to send these to us in the first place, however, different subscribers may want different filters applied and that could get ugly. 

If that ability is added in the api in the future, this can be removed or tweaked to tell the api what to deliver.

https://github.com/meteyou/mainsail/issues/51

image

Simple temperature filter:
`^(?:ok\s+)?(B|C|T\d*):`

I'm still green with vue and vuetify. Inside ConsolePanel.vue, if I did not use blur, it would update $store with every keypress. Since the expressions are cached when evaluated, I didn't want a race between messages coming in and compiling partial regular expressions.

Thinking about it, probably need a way to validate the filters are valid regular expressions, or in the filterMessage method, handle those that don't compile? This may need to be resolved before merging. It works how it is, but only the happy path.

### DIFF
--- a/src/components/panels/Settings/ConsolePanel.vue
+++ b/src/components/panels/Settings/ConsolePanel.vue
@@ -1,0 +1,59 @@
+<style>
+</style>
+
+<template>
+    <v-card>
+        <v-list-item>
+            <v-list-item-avatar color="grey"><v-icon dark>mdi-console-line</v-icon></v-list-item-avatar>
+            <v-list-item-content>
+                <v-list-item-title class="headline">Console</v-list-item-title>
+            </v-list-item-content>
+        </v-list-item>
+        <v-divider class="my-2"></v-divider>
+        <v-card-text class="px-0 pb-0 pt-3 content">
+            <v-row>
+                <v-col class="px-10 py-0">
+                    <v-textarea
+                        outlined
+                        name="input-7-4"
+                        label="Filters"
+                        v-model="filters"
+                        @blur="updateFilters($event)"
+                    ></v-textarea>
+                </v-col>
+            </v-row>
+        </v-card-text>
+    </v-card>
+</template>
+
+<script>
+    export default {
+        components: {
+
+        },
+        data: function() {
+            return {
+
+            }
+        },
+        computed: {
+            filters: {
+                get() {
+                    let c = this.$store.state.gui.console || {};
+                    return (c.filters || []).join('\n');
+                },
+                set() {
+                }
+            },
+        },
+        methods: {
+            updateFilters($event) {
+
+                let inputValue = $event.currentTarget.value.trim();
+                let filters = [...new Set(inputValue.split('\n'))];
+
+                return this.$store.dispatch('setSettings', { gui: { console: { filters: filters } } });
+            }
+        }
+    }
+</script>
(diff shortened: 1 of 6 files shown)


randellhodges:
I added some rudimentary validation. I am not sure how to tie the validation together with the button in interface.vue. Presently, it won't prevent the save, but it won't save invalid filters at least.

Also changed to not calculate the expressions on demand, but calculate them when the incoming state data has filters specified and is being stored. Not sure if this is better or not.

randellhodges:
@meteyou Any feedback?

src/components/panels/Settings/ConsolePanel.vue
```diff
@@ -0,0 +1,86 @@
...
+                    ></v-textarea>
+                </v-col>
+            </v-row>
+        </v-card-text>
+    </v-card>
+</template>
+
+<script>
+
+    function getFiltersFromText(text) {
```
meteyou, line 34: you can move this function into methods

src/components/panels/Settings/ConsolePanel.vue
```diff
@@ -0,0 +1,86 @@
...
+        let inputValue = text.trim();
+        let filters = [...new Set(inputValue.split('\n'))];
+        return filters;
+    }
+
+    function getFilterErrors(filters) {
```
meteyou, line 40: same as getFiltersFromText. you can move this function into methods

src/store/mutations.js
```diff
@@ -4,6 +4,29 @@ import { findDirectory } from '../plugins/helpers';
 import { colorArray, colorChamber, colorHeaterBed/*, temperaturChartSampleLength*/ } from './variables';
 import defaultPrinter from './printer';
 
+let consoleFilterCache = [];
+
+function updateConsoleFilterCache(filters) {
```
meteyou, line 9: you can move this function to plugins/helpers and only import these into the mutations

src/store/mutations.js
```diff
@@ -4,6 +4,29 @@ import { findDirectory } from '../plugins/helpers';
...
+                const expr = new RegExp(filter);
+                consoleFilterCache.push((t) => expr.test(t));
+            } catch {
+                console.error('Error compiling filter: ' + filter);
+            }
+        }
+    });
+}
+
+function filterConsoleMessage(text) {
```
meteyou, line 23: you can move this function to plugins/helpers and only import these into the mutations

randellhodges:
Thank you for in depth feedback. I'll get these resolved as soon as I can.

LazaroFilm:
What about just a "Show temperatures in the console" toggle button? I doubt we need a full RegEx filter available to the user. That would remove the need to check if the expression is good, simplify the code and streamline the user experience (they wouldn't need to copy/paste that RegEx themselves.

meteyou:
pls can you update your PR? if you need help with the new .env.development.example feel free to ping me. i will write a doc file in the next days.

jperrin72:
Hey, this is a great addition to mainsail, any chab-nce this could be merge ? thx

meteyou:
i added it by my own... https://github.com/meteyou/mainsail/commit/3c217e1161e093df90f8467067e932a428d44f2a

PR #153, opened 2020-12-29
feat(refresh-webcam): Refresh webcam panel on focus

shinuza:
Computers with agressive power management policies tend to stop the webcam feed after getting to sleep.

This forces the webcam feed to be restarted on focus.

### DIFF
--- a/src/components/panels/WebcamPanel.vue
+++ b/src/components/panels/WebcamPanel.vue
@@ -12,7 +12,7 @@
             </v-toolbar-title>
         </v-toolbar>
         <v-card-text class="px-0 py-0 content">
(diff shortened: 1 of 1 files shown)


shinuza:
Rebased to 07072ce

meteyou:
thank you for your pull request. this feature looks very promising. i only have 1 change in mind. can you please modify this, that the webcam url is not changed in store.gui.webcam.url? otherwise this will be saved with the .mainsail.json back to the host and the user would also see this change every time in the gui under settings > interface.

shinuza:
Sure, I'll look into it.

shinuza:
Rebased to 93c952c and code has been upload to maintain a local state instead of changing the url in the store.

meteyou:
thx! works really nice!

PR #167, opened 2021-01-17
Just simplifying nginx config in manual setup documentation

ru-ace:
Simplifying nginx config in manual setup documentation by replacing multiply locations entry with regex one.


meteyou:
Thx! works fine!

PR #266, opened 2021-04-20
fix power control panel

1ac3124:


meteyou:
what do you think you've fixed? can you explain the PR?

PR #299, opened 2021-06-22
Console QOL

freasy:
Added some console features. Multiline, linux like autocomplete and a guid helplist


freasy:
i will redo it

PR #312, opened 2021-07-20
Adding additional CSS classes for theming flexibility

StoneColdCrazy:
With these additional classes, creating your own themes is easier and can also be done in more detail. 
Adding additional CSS classes for more theming flexibility.


meteyou:
implemented with this PR: https://github.com/meteyou/mainsail/pull/372

PR #413, opened 2021-11-17
refactor: button overhaul and minor changes to the ui

steadyjaw:
```
* buttons change to match new design
* added autofocus to some dialogs
* added close on ESC to some dialogs
* changed some panelbars to new design
* extended throttled states info width in favor of readability
```


steadyjaw:
we may wait with a merge until everybody had the chance to check the changes made

PR #424, opened 2021-11-22
FR: increase scale for those w/ warped beds

vajonam:
increase scale from 0.5 to 2 for those with warped beds who can never get it any better!

### DIFF
--- a/src/pages/Heightmap.vue
+++ b/src/pages/Heightmap.vue
@@ -225,7 +225,6 @@ export default class PageHeightmap extends Mixins(BaseMixin) {
     private calibrateDialog = false
     private newName = ''
 
-    private heightmapScale = 0.5
     private probedOpacity = 1
     private meshOpacity = 1
     private flatOpacity = 0.5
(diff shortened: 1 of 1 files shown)


vajonam:
also added saving, found that this wans't being saved

meteyou:
this is only the "minimum" range. the min/max values automatically adjust to the mesh

vajonam:
this for the box scale, to allow it to go past 0.5 to 2. this way 0.3 variance doesn't look as bad

meteyou:
yes, because 0.3mm are really bad. thats why we use 0.5 as "min" box size to see anything.

vajonam:
No worries. You can close the PR. I can keep this on my local build.

vajonam:
@meteyou  what about the saving of the sate? that is possibly useful ? when a user selects the states to save it?

meteyou:
@vajonam i dont know what you mean. can you explain it?

vajonam:
The Scale slider position isn't' saved right now. every you you come back to page it gets reset. 

The PR include a bit to save that settings.

PR #426, opened 2021-11-22
fix: add theming for find/search panel

vajonam:
adding theme for search panel

### DIFF
--- a/src/components/TheEditor.vue
+++ b/src/components/TheEditor.vue
@@ -1,4 +1,57 @@
-<style scoped>
+<style lang="scss">
+
+*:focus:not(.focus-visible) {
+  outline: none;
+}
+
+[type=checkbox] {
+  width: 2.2em;
+  height: 2.2em;
+  color: dodgerblue;
+  vertical-align: middle;
+  -webkit-appearance: none;
+  background: none;
+  border: 10px;
+  outline: 0;
+  flex-grow: 0;
+  border-radius: 4px;
+  background-color: var(--v-toolbar-base);
+  transition: background 300ms;
+  cursor: pointer;
+}
+
+
+/* Pseudo element for check styling */
+
+[type=checkbox]::before {
+  content: "";
+  color: transparent;
+  display: block;
+  width: inherit;
+  height: inherit;
+  border-radius: inherit;
+  border: 2px;
+  background-color: transparent;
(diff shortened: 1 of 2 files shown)


vajonam:
image

The main text is set to primary colour and re-using other background variables.

vajonam:
updated LnF 
image

vajonam:
@steadyjaw  anything more you need from me on this?

steadyjaw:
awesome. thanks for your work.
looks cool. But I am still missing some styling.
image

Would be awesome, if you could dig a little bit deeper.

The following is not matching the Mainsail UI from my point of view
**General**
- needs a little bit more spacing (padding and margin on elements)

**Panel**
- background color -> maybe try "--v-panel-base" or "-v-toolbar-base"?

**Buttons**
- font weight too low
- more letter spacing 
- hover-effect -> no transition, wrong color
- if same as the panel color maybe they could be elevated or make them flat, we use both styles within the UI
Example:
image

**Close Button*
looks cool, I would have tried to add the mdi-close icon. But the way you did it, it looks ok
- no hover effect on the button, could you try to make it look like the other buttons with an icon only?

**text-areas**
looks totally different the the other
Example:
text-area

**Checkboxes**
same as above
Example:
checkboxes

Please let me know if you need something.
If some of the styling is too complex and it cannot be added to the codemirror config file, it is totally fine to add the styling to the "TheEditor" component in: https://github.com/mainsail-crew/mainsail/blob/develop/src/components/TheEditor.vue
There is a "styling" tag at the top. It says "scoped" and is empty. Maybe you have to remove the "scoped" prop.

steadyjaw:
@vajonam 
you interested in doing some additional changes for this PR?

vajonam:
> 

Sure, let me know what you more you want done?

vajonam:
@steadyjaw  what else do you need me to do here?

vajonam:
@th33xitus can you review this too :-) this is been sitting around forever

th33xitus:
@vajonam can you update this PR and resolve all conflicts please? we can then have a look at it. also lets see what @meteyou has to add to this maybe.

vajonam:
Sure. Will do this later today.

vajonam:
will send new PR for this.

PR #427, opened 2021-11-22
fix: round variance to 3 decimal places

vajonam:
fix for rounding variance to 2 decimal places.


vajonam:
image

meteyou:
sry... i fixed this also today...

PR #437, opened 2021-11-29
Fix macro param regexp

### LINKED ISSUE #436: Some macros do not report their parameters
**Describe the bug:**\
Some macros do not report requiring parameters in the interface
Eg:
```ini
[gcode_macro SET_BED_SOAK_TIME]
gcode:
  SAVE_VARIABLE VARIABLE=bed_soak_time VALUE={params.SECONDS|default(120)}
```

While some of them do:
```ini
[gcode_macro SET_INFILL_SQV]
gcode:
    {% set svv = printer.save_variables.variables %}
    {% set val = params.SQV|default(svv.infill_sqv|default(printer.configfile.settings.printer.square_corner_velocity, true), true) | int %}
    SAVE_VARIABLE VARIABLE=infill_sqv VALUE={val}
```

**Expected behavior:**\
The macro with parameters should report their parameters in the interface.

**Screenshots:**\
image

**Logfiles:**\
Please upload your Logfiles here or upload them to [pastebin.com](http://pastebin.com) and paste your log links here.
- klippy.log
- moonraker.log

RomRider:
Fix #436

### DIFF
--- a/src/plugins/helpers.ts
+++ b/src/plugins/helpers.ts
@@ -194,7 +194,7 @@ export function formatTime(date: Date): string {
 }
 
 export function getMacroParams(macro: { gcode: string }): PrinterStateMacroParams {
-    const paramRegex = /{%?.*?\sparams\.([A-Za-z_0-9]+)(?:\|(int|string|double))?(?:\|default\('?"?(.*?)"?'?\))?(?:\|(int|string))?.*?%?}/
+    const paramRegex = /{%?.*?\s*params\.([A-Za-z_0-9]+)(?:\|(int|string|double))?(?:\|default\('?"?(.*?)"?'?\))?(?:\|(int|string))?.*?%?}/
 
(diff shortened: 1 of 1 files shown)


freasy:
Thanks for the issue! But if there is no enforced whitespace `\s` is obsolet. So the correct regex for matching both cases (and the rest) is:
`/{%?.*?params\.([A-Za-z_0-9]+)(?:\|(int|string|double))?(?:\|default\('?"?(.*?)"?'?\))?(?:\|(int|string))?.*?%?}/`

There everything is covered by `.*?` since `\s*` is included in that term.

We will change the regex in future releases :)

RomRider:
I've updated it :) You are correct, there's no use for \s* anymore at all!

freasy (approved):

meteyou:
thx for the PR!

PR #441, opened 2021-11-30
fix: add icons for print settings

vajonam:
adding icons for the print settings.

image


Lanovar:
As much as I would like to see the rabbit in Mainsail 😜, we discussed this internally and decided it wouldn‘t fit the new „lock slider“ feature added in 2.1beta. Too many icons in the same place.

vajonam:
So then are you removing the icons for the other sliders that already have icons? I agree on a touch device it is cluttered, but on a desktop it looks nice.

vajonam:
This is what they all look like with all the locking sliders.

image

PR #465, opened 2021-12-10
feat: add min_power support for fans (Klipper feature #4720)

HairingX:
The Klipper feature #4720 is not merged in yet, but seems to be within a short time.
https://github.com/Klipper3d/klipper/pull/4720

This feature request supports the new min_power fan property feature when it arrives.

It should work just as today unless the min_power fan property is present in klipper config. 
Therefore it can be merged today without issues, but is only needed when the Klipper feature request is merged.

### DIFF
--- a/src/components/inputs/MiscellaneousSlider.vue
+++ b/src/components/inputs/MiscellaneousSlider.vue
@@ -77,10 +77,10 @@ export default class MiscellaneousSlider extends Mixins(BaseMixin) {
     convertName = convertName
     private timeout: number | undefined
     private isLocked = false
-    private min = 0
     private value = 0
 
     @Prop({ type: Number, required: true }) target!: number
+    @Prop({ type: Number, default: 0 }) min!: number
     @Prop({ type: Number, default: 1 }) max!: number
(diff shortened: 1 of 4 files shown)


th33xitus (changes requested):
Indentation level is 4 spaces each.
Please stick to that formatting and fix the whitespace changes in `src/components/inputs/MiscellaneousSlider.vue` from line 2 to line 65 and in line 166

HairingX:
Gosh', the whitespace/newline comparison was toggled off in VS Code git changes.

Thank you for pointing it out, indentation issues are reverted to the correct indentation.

th33xitus:
@HairingX Hey! I just wanted to get back to this PR and see how the current state is (apart from it conflicting with the current development state of Mainsail but that is no real issue).

I saw that your PR for Klipper, which this PR relays on, got recently closed: https://github.com/Klipper3d/klipper/pull/4720
Do you intend to resurrect the Klipper PR? Otherwise the PR for Mainsail would sadly become obsolete i guess.

HairingX:
I have closed it. We could not agree on the implementation :-)

PR #506, opened 2022-01-03
docs: Update First Boot docs

prcutler:
Update the First Boot page Update section based on review feedback.  Also added additional changes in the opening First Boot paragraphs that will need to be reviewed.

I merged develop into my develop before starting - I don't know why it's showing the previous commits that have already been merged (or how to fix) - I'm so sorry.


lixxbox:
I adjusted some codeblocks.
Apart from that, it looks very good! Thanks a lot!

This would probably render PR #469 obsolete and it could be closed, right?

prcutler:
Yes, it should.  Thanks for the review and I'll make sure to do the codeblocks that way going forward.

PR #539, opened 2022-01-17
fix: encodeURI for thumbnails and timelapse files

ngynkvn:
GET requests including file names with special characters such as '%' will properly encode the request to the server.

## Description

I run mainsail and moonraker behind Nginx. My g-code files are named to include infill percentage, i.e. "25%", which for URLs is a reserved escape character. Nginx parses these requests incorrectly and returns a 400 Bad Request code, leaving thumbnail icons and timelapse videos inaccessible for me.

### Before fix:
Before

Clicking the first file in timelapse view breaks:

Before2

Note the g-code file without the '%' symbol successfully obtains it's thumbnail, but all others are breaking.

I did a grep through the code base and replaced lines where encodeURI was missing for the GET request.

### After fix:

Working2

The first timelapse video is successfully found and displayed:

working-vid

Lmk about any concerns, and thank you for providing a great and useful UI! 🎉

### DIFF
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -924,7 +924,9 @@ export default class GcodefilesPanel extends Mixins(BaseMixin) {
                 thumb.height >= thumbnailSmallMin && thumb.height <= thumbnailSmallMax
             )
 
-            if (thumbnail && 'relative_path' in thumbnail) return this.apiUrl+'/server/files/'+this.currentPath+'/'+encodeURI(thumbnail.relative_path)+'?timestamp='+item.modified.getTime()
+            if (thumbnail && 'relative_path' in thumbnail) {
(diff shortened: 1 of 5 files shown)


meteyou:
thx for the PR!

wlhlm:
Heads up: Funneling the whole URL through `encodeURI()` will most likely break for users on IPv6 as can be seen in the following fluidd PR: https://github.com/fluidd-core/fluidd/pull/439

meteyou:
> 



> Heads up: Funneling the whole URL through `encodeURI()` will most likely break for users on IPv6 as can be seen in the following fluidd PR: [fluidd-core/fluidd#439](https://github.com/fluidd-core/fluidd/pull/439)

thx for your feedback. should be fixed with this commit: https://github.com/mainsail-crew/mainsail/commit/56d61ce6ae9af82e4e97d86f9e418b214a6b2fc5

PR #540, opened 2022-01-17
fix: set default for min_extrude_temp

r3Fuze:
`min_extrude_temp` is undefined if the user has not specified `min_extrude_temp` in their Klipper config leading to a small display error in the extrude/retract tooltip:
Screenshot 2022-01-17 224253

This PR sets `min_extrude_temp`to Klipper's default 170 if it is undefined.

Signed-off-by: Morten Lindhardt <r3fuze@gmail.com>

### DIFF
--- a/src/components/panels/ControlPanelExtruder.vue
+++ b/src/components/panels/ControlPanelExtruder.vue
@@ -121,7 +121,7 @@ export default class ControlPanelExtruder extends Mixins(BaseMixin) {
     }
 
     get minExtrudeTemp() {
-        return this.$store.state.printer.configfile.config.extruder.min_extrude_temp ?? 170
+        return this.$store.state.printer.configfile.settings.extruder.min_extrude_temp
     }
 
     doSend(gcode: string) {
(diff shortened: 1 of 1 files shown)


src/components/panels/ControlPanelExtruder.vue
```diff
@@ -121,7 +121,7 @@ export default class ControlPanelExtruder extends Mixins(BaseMixin) {
     }
 
     get minExtrudeTemp() {
-        return this.$store.state.printer.configfile.config.extruder.min_extrude_temp
+        return this.$store.state.printer.configfile.config.extruder.min_extrude_temp ?? 170
```
meteyou, line 124: pls use printer.configfile.settings instead of config.

config = user config file
settings = parsed and filled with all default settings

meteyou (changes requested):

src/components/panels/ControlPanelExtruder.vue
r3Fuze: Thanks, fixed

meteyou (approved):

PR #590, opened 2022-01-28
chore: remove unused packages

pataar:
These packages aren't in use, and can be removed as dependency safely.


pataar:
Will be integrated in #594

PR #594, opened 2022-01-28
build(vite): migrate to vitejs

pataar:
Migrating from https://cli.vuejs.org/ to https://vitejs.dev/guide/why.html.

You can try this out by cloning the fork and running `docker-compose up --build`


meteyou (approved):

src/store/socket/index.ts
```diff
@@ -7,11 +7,11 @@ import {RootState} from '@/store/types'
 
 export const getDefaultState = (): SocketState => {
     return {
-        remoteMode: process.env.VUE_APP_REMOTE_MODE?.toLowerCase() === 'true' || (document.location.hostname === 'my.mainsail.xyz'),
-        hostname: process.env.VUE_APP_HOSTNAME || (process.env.VUE_APP_REMOTE_MODE?.toLowerCase() === 'true' || document.location.hostname === 'my.mainsail.xyz' ? '' : window.location.hostname),
-        port: process.env.VUE_APP_PORT || (process.env.VUE_APP_REMOTE_MODE?.toLowerCase() === 'true' || document.location.hostname === 'my.mainsail.xyz' ? 7125 : window.location.port),
+        remoteMode: import.meta.env.VUE_APP_REMOTE_MODE === true || (document.location.hostname === 'my.mainsail.xyz'),
+        hostname: (import.meta.env.VUE_APP_HOSTNAME as string) || ((import.meta.env.VUE_APP_REMOTE_MODE === true || document.location.hostname === 'my.mainsail.xyz') ? '' : window.location.hostname),
+        port: Number(import.meta.env.VUE_APP_PORT || (import.meta.env.VUE_APP_REMOTE_MODE === true || document.location.hostname === 'my.mainsail.xyz' ? 7125 : window.location.port)),
```
wlhlm, line 12: Mainsail has been trying to connect to moonraker on port 0 (rather than port 80) ever since this PR has been merged. This issue is that in the case Mainsail falls back onto `window.port.location`, if the browser has loaded mainsail via HTTP on port 80, said variable will be an empty string (as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Location/port)). `Number('')` yields 0, which makes Mainsail connect on the wrong port. Removing the cast to number fixes the issue.

pataar: Good finding! Typescript doesn't like strings though. Perhaps undefined should work. I can make a PR if you want?

wlhlm: Please do. I have pretty much no js/ts experience to be able to submit a competent fix.

wlhlm: I think the important thing is that we generate the correct URL here: https://github.com/mainsail-crew/mainsail/blob/0cf8ba05be1a3e43201b94914ab150efe4562a8c/src/store/socket/getters.ts#L8

pataar: Fixed in #631

PR #610, opened 2022-02-06
fix: properly size Gcode drag zone in Safari

rufo:
In Safari, the drag zone for Gcode doesn't show up properly, making it really hard to drop Gcode files into Mainsail:

https://user-images.githubusercontent.com/2745/152683919-aaaef124-a70d-413d-8dc1-9cde7ab4e61d.mov

I did some Googling, and [found a suggestion on Stack Overflow to try using `100vh` instead of `100%`](https://stackoverflow.com/a/44819258), which seems to work:

https://user-images.githubusercontent.com/2745/152683977-e66e3f60-f7e6-4d14-b918-b1dc42fd312d.mov

It seems to look the same in Firefox/Chrome to me now, though I understand if there are any concerns from a random CSS change I pulled off Stack Overflow 😂

### DIFF
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -19,7 +19,7 @@
     display: table;
     position: absolute;
     top: 0; left: 0;
-    width: 100%; height: 100%;
+    width: 100%; height: 100vh;
     z-index: 9999999999;
     background-color: rgba(0, 0, 0, 0.5);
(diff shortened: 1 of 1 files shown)


meteyou:
hey! thx for your PR. i fixed it a little bit different. 100vh would be to big.

this is my fix for this issue. i will publish a small bugfix release
https://github.com/mainsail-crew/mainsail/commit/1fa705afbb926d4504ca67eccdf9aa168c3aee24

PR #612, opened 2022-02-07
fix(editor): partial improvement of config syntax highlighting

freasy:


th33xitus (changes requested):
- failed checks
- empty functions
- unused code imports
- invalid character for comment in config defined
- invalid character for comment in gcode defined

meteyou (approved):

PR #635, opened 2022-02-14
ci: run build size report on 'analyze' label

pataar:
The size report is only ran after adding the label. This way `pull_request_target` is still safe to use.

Like this:

image

If you want something different than `analyze`, just let me know.

Also updated the linter jobs to be a bit more standard.


th33xitus:
Imho `analyze` is fine. I went ahead and already created that label.

Issue #404, opened 2021-11-05
[FR] Export Print History as .CSV


troypouliot:
**Is your feature request related to a problem? Please describe.**
Yes, I keep a single excel log for all of my printers to track total hours. Every month or every other month I will export a .csv file from the Octoprint plugin "Print Job History". This exports a new row for each print job with all the relevant info (pretty much everything listed in the Mainsail print history).
This makes it a breeze to copy/paste the needed rows to my log file to update the cumulative printer time.

**Describe the solution you'd like**
A button to export as .CSV all or selected print jobs with all relevant info in the print history tab.

**Describe alternatives you've considered**
Going back to Octoprint ;(

**Additional context**
This very important feature is also missing from the Fluidd platform. I chose to request it here as this project looks to be a bit more active with development.

Thanks!

![image](https://user-images.githubusercontent.com/57006528/140440455-adb1d8f2-7637-452d-a605-b6730e585b67.png)

meteyou:
thx for your request. can you upload an example with 1-2 jobs in the CSV file? this feature should not be a problem to implement it.

troypouliot:
Here you go-

[OctoprintPrintJobHistory (12).csv](https://github.com/meteyou/mainsail/files/7486513/OctoprintPrintJobHistory.12.csv)

Not all columns are required for me, but other may find use for them. 

My personal record keeping log uses the columns;
1. Print Result - Used to help filter out failed prints (I also look for print time to be < 10 mins)
2. Start Datetime - for sorting
3. Duration - Obvious :)

Thanks for the quick response! Keep up the great work.

meteyou:
np. how do you sort your printers? would it be nice to have one column with the printername?

troypouliot:
It wouldn't hurt, but I don't know that it is necessary either.
Right now, each printer runs it's own Octoprint instance, so when it is time to log hours I just end up with a seperate csv file for each printer and handle them individually. 

I only have one printer running Mainsail right now as I test it out and make sure it has what I need before switching over the others.

So, yeah now that I think about it more, it would be useful to have the printer name _somewhere_. Either a column that includes this. 
Or, since I think I will be doing the same thing with running seperate instances of Mainsail for each printer, maybe include the printer name in the filename of the csv export? something like [printer_name]-print_history-[export date].csv.
That would make it easy to keep all of them together in a directory.

meteyou:
[...]

PR #675, opened 2022-02-24
feat: export print history as CSV

meteyou:
This PR add a function to export selected or all jobs to an CSV file.

fix: #404 


wlhlm:
I see this just joins the strings of each job field with a semi-colon. Does that work safely with filenames that might contain reserved characters?

pataar:
Note that the delimiter is different in the US for example. EU: ';', USA/UK: ','

meteyou:
> Does that work safely with filenames that might contain reserved characters?

i think that you would have problems with uploading or printing and that's why you wouldn't even see something like that in the history.

meteyou:
> Note that the delimiter is different in the US and UK. EU: ';', USA/UK: ','

ohhh... thx for this info! because i just use US number format for the export, i think it will be fine to use ',' only.

wlhlm:
> > Does that work safely with filenames that might contain reserved characters?
> 
> i think that you would have problems with uploading or printing and that's why you wouldn't even see something like that in the history.

I have just tested this and no, you can't currently print a with a semicolon in the name because Klipper treats everything following a semi-colon as GCode comment regardless of whether it is quoted or not.

However, he Klipper stack certainly works with commas in filenames as I've been using such a filename scheme for ages and that will result in broken CSV output when using a mainsail language that results in the comma field-separator as you've just implemented.

meteyou:
@wlhlm thx for testing. i pushed a fix. with the new commit i use the local number format and escape the csv seperator in strings. pls test this again, if this would fit.

wlhlm:
Thx for the fix! I'm wondering why not pull in a battle-tested CSV generator as a dependency that does proper escaping instead of a home-grown implementation? Just skimming through my mainsail node_modules, I could find [d3-dsv](https://github.com/d3/d3-dsv) which is an indirect dependency of mainsail. It's a pretty straight-forward module with about 160LOC.

meteyou:
Because it's easier and faster to generate it this way and I don't want to add new deps for it (the module you linked above is to parse a CSV file. Reverse direction I need in this case).

wlhlm:
> (the module you linked above, I think it is to parse a CSV file. Reverse direction I need in this case).

Have a look  at [`d3.csvFormat()`](https://github.com/d3/d3-dsv#csvFormat).

meteyou:
> Have a look at [d3.csvFormat()](https://github.com/d3/d3-dsv#csvFormat).

this function do complete the same as i implemented. i have to define the seperator exactly the same way... its more complex to use a lib for this very very small usage. a CSV file is just a text file. if we would export an excel file, i would use a dep.

PR #676, opened 2022-02-25
fix: removing remote printer in remote mode

meteyou:


th33xitus (approved):
Tested and works!

PR #702, opened 2022-03-06
style: order all locale keys alphabetically

pataar:
This will improve the ease of reviewing and reduces the chance of conflicts.

While I was at it, I also update both the dependencies and some Github action versions.


package.json
```diff
@@ -12,8 +12,8 @@
         "format": "npm run format:base -- --write",
         "format:base": "prettier .",
         "format:check": "npm run format:base -- --check",
-        "lint": "eslint --ext .ts,.vue --ignore-path .gitignore src",
-        "lint:fix": "eslint --ext .ts,.vue --ignore-path .gitignore src --fix",
+        "lint": "eslint --ignore-path .gitignore src",
+        "lint:fix": "npm run lint -- --fix",
```
meteyou, line 16: is there a `--` too much?

pataar: ` -- --fix` basically means: pass down `--fix` to the command the command itself is calling. So it passes down the arguments to the next one.

meteyou (approved):

PR #714, opened 2022-03-13
fix: add missing translation keys

### LINKED ISSUE #510: [FR] missing translation keys
While going through the translations for Klipperscreen, and cross-referencing to make them similar, I stumbled across these minor parts missing from the translator files. (Or i can't find them)

All text on the "History-wheel" is not available for translations
image

Estimation method selectors not available for translation
Filament and Slicer are the same, but "File" is supposed to be "Fil"
image

"Root" which I'm not sure i see the need for is also nowhere to be found
image

meteyou:
- estimate settings in interface setting
- "root" dropdown in config file panel

this PR fix #510 

### DIFF
--- a/src/components/panels/Machine/ConfigFilesPanel.vue
+++ b/src/components/panels/Machine/ConfigFilesPanel.vue
@@ -12,7 +12,7 @@
                         <v-select
                             v-model="root"
                             :items="registeredDirectories"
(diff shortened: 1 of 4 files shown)


th33xitus (dismissed):
LGTM!

th33xitus (approved):

PR #715, opened 2022-03-13
style: indent size of 4 spaces in json

th33xitus:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,10 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{json,yml}]
+[*.json]
 indent_size = 4
+indent_style = space
+
+[*.yml]
+indent_size = 2
 indent_style = space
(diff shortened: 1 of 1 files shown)


pataar:
I'd say, keep `yml` on two spaces. This is generally used as standard

th33xitus:
> I'd say, keep `yml` on two spaces. This is generally used as standard

Im fine with that. So we should do it rather like this and split json and yaml?
```yaml
[*.json]
indent_size = 4
indent_style = space

[*.yml]
indent_size = 2
indent_style = space
```

pataar:
> > I'd say, keep `yml` on two spaces. This is generally used as standard
> 
> Im fine with that. So we should do it rather like this and split json and yaml?
> 
> ```yaml
> [*.json]
> indent_size = 4
> indent_style = space
> 
> [*.yml]
> indent_size = 2
> indent_style = space
> ```

Perfect

pataar (approved):

meteyou (approved):

PR #734, opened 2022-03-26
fix: check if panel exists before load on dashboard

meteyou:


src/pages/Dashboard.vue
```diff
@@ -97,6 +97,7 @@ import StatusPanel from '@/components/panels/StatusPanel.vue'
 import ToolsPanel from '@/components/panels/ToolsPanel.vue'
 import WebcamPanel from '@/components/panels/WebcamPanel.vue'
 import ZoffsetPanel from '@/components/panels/ZoffsetPanel.vue'
+import _ from 'lodash'
```
pataar, line 100: Due to bundle size, it's better if we'd use `lodash.kebabcase` as package. This is much smaller than the full `lodash` package.

meteyou: thx and fixed it!

pataar (approved):

PR #750, opened 2022-04-10
fix: resize tempchart on window resize event

meteyou:


th33xitus:
While switching through different resolutions in the chrome dev tools, this error regularly appears.
The warning was already present before though, would be good if it is somehow possible to fix that as well.

image

th33xitus (changes requested):
Required: Fix console errors
Optional: Fix console warnings

th33xitus (approved):

PR #754, opened 2022-04-16
fix: add error message in webcam panel, if no webcam is available

### LINKED ISSUE #752: Webcam site display no configuration available message
### Requested feature:

When users (including me) try to add their webcam, expect that the cam is automatically shown on the webcam site.
But even if the users get the [crowsnest](https://github.com/mainsail-crew/crowsnest) or the old webcam service running, is the site webcam empty. Only the gray box is shown. So sometimes is it not obviously that the webcam needs also configured under the settings.

So Things would be good.
- A message where you write in the empty box, "Add your webcam under the Interface Settings -> Webcams." That `div` is anyway replaced by the cameras if one is configured, so you don't need to write a complex logic for that. (hotfix possible!?)

- A add button that open the Right Webcams settings menu. (Maybe obsolet but maybe also a good option)

- Automatic add default settings when crowsnest detect a capable webcam. (maybe with the settings that configured in the webcam.conf) (only added when no configuration exist) + a settings button under the webcams on the webcam site not on the dashboard

confused users. 
Difficult to find documentation (makes documentation slightly obsolet) (the best documentation is the one that is not needed.)
Discord inquiries and waste of time. usability

### Additional information:

_No response_

meteyou:
This PR fix #752 

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -593,6 +593,7 @@
             "All": "All",
             "FPS": "FPS",
             "Headline": "Webcam",
+            "NoWebcam": "No webcam available. Add your webcam under the \"Interface Settings\" -> \"Webcams\".",
             "UnknownWebcamService": "Unknown Webcam Service",
             "UrlNotAvailable": "URL not available"
         },
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -593,7 +593,7 @@
             "All": "All",
             "FPS": "FPS",
             "Headline": "Webcam",
-            "NoWebcam": "No webcam available. Add your webcam under the \"Interface Settings\" -> \"Webcams\".",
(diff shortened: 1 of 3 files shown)


src/locales/en.json
```diff
@@ -593,6 +593,7 @@
             "All": "All",
             "FPS": "FPS",
             "Headline": "Webcam",
+            "NoWebcam": "No webcam available. Add your webcam under the \"Interface Settings\" -> \"Webcams\".",
```
th33xitus, line 596:
```suggestion
            "NoWebcam": "No webcam available. Add a webcam under \"Interface Settings\" -> \"Webcams\".",
```

th33xitus (changes requested):
see suggestion

th33xitus (approved):

PR #760, opened 2022-04-22
Create sv-SE.json

eltacoloco:
Added Swedish translation


th33xitus:
Please do not commit IDE specific settings files. The settings.json and extensions.json do not belong into this PR.
On another note, it would be good to run Prettier on thesv-SE.json for a correct formatting of that file.
Thanks in advance!

PR #763, opened 2022-04-24
fix: hide unknown panels in interface settings > dashboard page

meteyou:


th33xitus (changes requested):
failed eslint checks

th33xitus (approved):

PR #777, opened 2022-05-06
feat: global fullscreen fileupload

meteyou:


src/locales/en.json
```diff
@@ -250,6 +253,7 @@
         "Later": "Later",
         "Mesh": "Mesh",
         "Name": "Name",
+
```
th33xitus, line 256: remove that blank line

th33xitus (changes requested):
please also add the german translation

th33xitus (approved):

PR #787, opened 2022-05-10
feat: add arm64 docker image support

samwiseg0:
Added arm64 image support to publish_docker.yml workflow file.

I have built and tested this workflow on my branch. Unfortunately, I could not get armv7 support to work. It seems that cypress does not support arm 32 bit. https://github.com/cypress-io/cypress/issues/4478 For my purposes I only needed an arm64 image.


th33xitus:
Thanks for the contribution!
I pushed a commit to this PR to make the Prettier test succeed, so we can merge it.

th33xitus (approved):

samwiseg0:
> Thanks for the contribution! I pushed a commit to this PR to make the Prettier test succeed, so we can merge it.

Thank you!

PR #826, opened 2022-05-18
fix: add file from sub directory to job queue

### LINKED ISSUE #825: [v2.2.0-beta3] gcodes in subfolder from gcode page not added to jobqueue
### Browser:

Chrome

### Device:

Desktop PC

### Operating System:

Windows

### What happened?

It is not possible to add gcode files from the gcode page to the jobqueue anymore, if the gcode is located in a subfolder.
Adding the same file in the same subfolder from the files tab in the status panel on the dashboard works fine.

After trying to add a gcode to the jobqueue this way and switching between dashboard and gcode file page also results in an issue in the files tab of the status panel:
22-05-18_11-00-59_chrome
It seems like all files from within the subfolder get added to the files tab, but with their location being the root directory. The files do not exist at that location. Trying to add those "phantom files" to the jobqueue or start printing them won't work obviously, they shouldn't be added to the files tab at all.

Correctly adding the selected file to the jobqueue.
Not causing an issue of displaying files that don't exist.

reproducing the initial issue:
1. switch to the gcode files page
2. switch into a subfolder
3. try to add a file to the jobqueue
4. observe that nothing seems to happen and the jobqueue panel is not appearing on the right side on the gcode files page
5. switch to the dashboard status panel
6. observe that no file was added to the jobqueue there as well

reproducing the displaying issue:
1. switch back to the gcode files page
2. switch back to the dashboard status panel files tab
3. observe that files were added to the list, that usually were located in the subfolder, now appear to be located in the root directory

### Additional information:

_No response_

meteyou:
fixes #825

### DIFF
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -1152,9 +1152,8 @@ export default class GcodefilesPanel extends Mixins(BaseMixin) {
     }
 
     addToQueue(item: FileStateGcodefile | FileStateFile) {
-        let path = this.currentPath.slice(7)
-        if (path != '') path += '/'
-        const filename = path + item.filename
+        let filename = [this.currentPath, item.filename].join('/')
+        if (filename.startsWith('/')) filename = filename.slice(1)
 
         this.$store.dispatch('server/jobQueue/addToQueue', [filename])
     }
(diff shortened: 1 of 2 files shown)


th33xitus:
The first part of the bug seems to be fixed. The second part not.
There still appear those "duplicates" after switching back and forth from dashboard to gcode files
image

th33xitus (approved):

Issue #845, opened 2022-05-26
impossible to set pressure advance values for extruder_stepper


dw-0:
### Browser:

Chrome

### Device:

Desktop PC

### Operating System:

Windows

### What happened?

Currently it is impossible to safely set pressure advance values for printer configurations where multiple extruders are configured with `[extruder_stepper]` sections in the printer.cfg. Using Mainsails input field for pressure advance will fail, as it tries to assign those pressure advance values to an `extruder`, that is not allowed to take pressure advance values. 
![22-05-26_14-16-52_chrome](https://user-images.githubusercontent.com/31533186/170487582-a6edcd1f-f2ce-4d5e-8053-2e30cbfd870d.png)
![22-05-26_14-16-34_chrome](https://user-images.githubusercontent.com/31533186/170487595-fbfff6e4-bab3-480f-b64b-299d6d530a26.png)

Even though an `extruder stepper` might be synced with an `extruder`, `extruder1`, `extruder2` and so on, it is currently impossible to figure that out which `extruder stepper` is synced with which `extruder`. That is due to technical limitations of Klipper, as Klipper (+Moonraker) does not offer a safe way to figure out which `extruder stepper` is currently the active one.

`[extruder]` is not allowed to have pressure advance values configured **and** won't accept any pressure advance values send to it via g-code. Klipper in turn requires the pressure advance to be set via g-code **and** by specifying the correct `extruder stepper` name, rather than the `extruder` name it is synced to.

After joint consultation with @pedrolamas (Fluidd developer) on how to tackle that issue we came to the following conclusion:
We will be unable to allow setting pressure advance values via input fields as long as Klipper offers a suitable solution to the problem above. For the meantime it is probably the safest solution to hide those input values once it's clear that there are `[extruder_stepper]` defined in the users printer.cfg. We have to wait for Klipper upstream changes to happen. Only then we can re-enable those input fields again.

Also see: https://github.com/fluidd-core/fluidd/issues/681

jalanjarosz:
Not sure if this info would help with creating a solution.  Here is a complete config setup of a multi-stepper extruder using extruder_stepper:

https://klipper.discourse.group/t/x-in-1-out-non-mixing-extruder-config/2387

Pressure Advance is set via the following command:

SET_PRESSURE_ADVANCE ADVANCE=0 EXTRUDER=[tool_name]

th33xitus:
[...]

PR #846, opened 2022-05-26
refactor: hide PA input fields if extruder_stepper is configured

th33xitus:
This PR is (hopefully only) a temporary solution to this issue: https://github.com/mainsail-crew/mainsail/issues/845
There is no safe way to set pressure advance values via those input fields, so we now disable them once we come across such printer configurations.

### DIFF
--- a/src/components/panels/ExtruderControlPanel.vue
+++ b/src/components/panels/ExtruderControlPanel.vue
@@ -146,8 +146,10 @@
                         attribute-name="S"></tool-slider>
                 </v-container>
                 <!-- PRESSURE ADVANCE SETTINGS -->
-                <v-divider></v-divider>
-                <pressure-advance-settings></pressure-advance-settings>
+                <template v-if="!extruderSteppers.length > 0">
+                    <v-divider></v-divider>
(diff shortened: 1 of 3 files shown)


danorder:
Next time please remember to add things like this to the stable release description. for the whats changed. almost went down a rabbit hole here seeing as fluidd still had p.a visible.

meteyou:
> Next time please remember to add things like this to the stable release description. Almost went down a rabbit hole here seeing as fluidd still had p.a visible. Thankfully i came across this thread first.

Pls read the issue #845 . Fluidd hide it in the same case!

danorder:
looking I have not updated to that version yet so that is on me so that would be why my bad.  Regardless that wasn't listed under the release personally I keep one "out of date" and the other more current as a fall back.

th33xitus:
This PR is listed in the changelog of the 2.2.0 stable release, as it appeared there first. It is not further mentioned in the 2.2.1 release notes. You find it under the "Bugfixes and Improvements" header, prefixed with "refactor". It is the very first one of the refactor changes. Although, i agree that we could have made it a bit more obvious, as the changelog got really long again this time and it seems it can be easily missed.

PR #861, opened 2022-05-29
locale(nl): update dutch localization

bartlammers:
Some empty strings still in there, but those were empty in the en.json as well so kept them as such.


th33xitus:
I pushed a fix with replacing the tabs with spaces as we require them.
There are a lot of empty strings in that file. So a lot of missing translations.
In the current english locale file are only 2 emptry strings. I this one are around 44.

th33xitus (changes requested):
I pushed a fix with replacing the tabs with spaces as we require them.
There are a lot of empty strings in that file. So a lot of missing translations.
In the current english locale file are only 2 emptry strings. I this one are around 44.

bartlammers:
Updated the translations, removed the empty strings (was because of BabelEdit, they apparently still existed in other languages) and re-ran lint and prettier

th33xitus:
Thank you, looks good to me now!

th33xitus (approved):

PR #874, opened 2022-06-08
locale(ko-kr): add new lanquage pack such that south korean users

ZISQO:
hi there! i've been transulate json file of english while browsing each menu, but i found that an instance required for translation was missing on recently mainsail version, but it was not added randomly when adding the first time work. i will check it again after you merging over this file.


meteyou:
Thank you for the language file. This comes just at the right time. Are you also on our discord server? Would love to add you to the translators group.

PR #896, opened 2022-06-12
locale(zh): update Chinese localization

Samuel-0-0:


Samuel-0-0:
locale(zh): update Chinese localization

meteyou (approved):

PR #932, opened 2022-06-23
style: Change emergency stop icon to avoid confusion on mobile

LDPE:
_Note: I didn't bother bringing this up anywhere before submitting this PR. I figured it was a quick enough edit that I wasn't out much time if the dev team doesn't want to make this change. Just consider this PR a suggestion._ 

Currently, the emergency stop icon is a circle with a exclamation mark inside. This can cause confusion on mobile when the "emergency stop" text is not displayed along side the icon. This is especially true for new users coming from octoprint, because octoprint uses an exclamation mark in the same location to alert users to warnings or errors detected. This confusion can cause new users to accidentally e-stop in the middle of a print thinking that they are just viewing an alert (ask me how I know). This has been brought up at least once on discord as well.

This PR changes the icon to an octagon with an exclamation mark inside. This "stop sign" should be a more intuitive indication of button function on mobile when the "emergency stop" text is hidden. Specifically, it changes the icon from `mdiAlertCircleOutline` to `mdiAlertOctagon`.

### DIFF
--- a/src/components/TheTopbar.vue
+++ b/src/components/TheTopbar.vue
@@ -91,7 +91,7 @@
                 class="button-min-width-auto px-3 emergency-button"
                 :loading="loadings.includes('topbarEmergencyStop')"
                 @click="btnEmergencyStop">
-                <v-icon class="mr-md-2">{{ mdiAlertCircleOutline }}</v-icon>
+                <v-icon class="mr-md-2">{{ mdiAlertOctagon }}</v-icon>
                 <span class="d-none d-md-inline">{{ $t('App.TopBar.EmergencyStop') }}</span>
             </v-btn>
             <the-notification-menu></the-notification-menu>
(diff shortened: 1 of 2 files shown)


meteyou:
Pls post a screenshot from the old icon and from the new icon in mainsail.

src/components/TheTopbar.vue
```diff
@@ -115,7 +115,7 @@
                 :title="$t('EmergencyStopDialog.EmergencyStop')"
                 toolbar-color="error"
                 card-class="emergency-stop-dialog"
-                :icon="mdiAlertCircleOutline"
+                :icon="mdiAlertOctagon"
```
th33xitus, line 118: You're missing the import for that icon.

th33xitus (commented):
First things first: You haven't checked the code you try to contribute. Your change doesn't work.

**Personally** i don't like the icon-type you choose. Though i probably would be fine if `mdiAlertOctagonOutline` was used.

Your reasoning about confusing new users that come from OctoPrint might be true. But Mainsail is not OctoPrint and the way i personally see it is, that a new interface also requires new things to learn. I can't imagine that new Mainsail users open the interface only on mobile devices for the very first time they use it and have never seen the actual description of the button. Apart from that, Mainsail now handles displaying notifications differently than before and make them accessible from the bell-icon. This is probably also something the user has to learn when coming from OctoPrint.

PR #933, opened 2022-06-25
feat: support tvoc units

nknotts:
Screen Shot 2022-06-24 at 8 28 26 PM

In the Voron community, we are developing a controller for the [nevermore max](https://github.com/nevermore3d/Nevermore_Max). Part of this will be displaying [VOC](https://en.wikipedia.org/wiki/Volatile_organic_compound) levels from sensors such as the [SGP30](https://www.adafruit.com/product/3709).

This is only a partial solution. Currently, the additional values are only displayed if the sensor is advertised as being in the list in [additionalSensors](https://github.com/mainsail-crew/mainsail/blob/6661e031fbe8d4662fa2c838ede6668f137a1dcb/src/store/variables.ts#L17). In my testing, I am advertising to be a `bme280`. We might want to think about making this more generic in the future. 

However, I think my change in its current form is noninvasive and self-contained.

### DIFF
--- a/src/store/printer/getters.ts
+++ b/src/store/printer/getters.ts
@@ -442,6 +442,10 @@ export const getters: GetterTree<PrinterState, RootState> = {
                                 tmp[key]['unit'] = '%'
                                 break
 
+                            case 'tvoc':
+                                tmp[key]['unit'] = 'ppb'
+                                break
+
                             default:
                                 tmp[key]['unit'] = ''
                         }
--- a/src/store/printer/getters.ts
+++ b/src/store/printer/getters.ts
@@ -446,6 +446,10 @@ export const getters: GetterTree<PrinterState, RootState> = {
                                 tmp[key]['unit'] = 'ppb'
                                 break
 
+                            case 'eco2':
+                                tmp[key]['unit'] = 'ppm'
+                                break
+
                             default:
                                 tmp[key]['unit'] = ''
                         }
--- a/src/store/variables.ts
+++ b/src/store/variables.ts
@@ -14,7 +14,7 @@ export const themeDir = '.theme'
 export const datasetInterval = 1000
 export const datasetTypes = ['temperature', 'target', 'power', 'speed']
 export const datasetTypesInPercents = ['power', 'speed']
-export const additionalSensors = ['bme280', 'htu21d']
+export const additionalSensors = ['bme280', 'htu21d', 'sgp30']
 
 /*
  * List of valid gcode file extensions


meteyou:
Nice! Thx! Can you link the Klipper branch for the SGP30. I have 2 sensors here for testing this PR.

nknotts:
For the initial design of the nevermore max controller firmware, the plan is to make it "self contained", as in, the controller will work even when not connected to klipper. Because of this, the sgp30 measurements will occur within the firmware. That is to say, I do not have an sgp30 klipper module at the moment. 

What I do have is a custom [klipper module for the nevermore max firmware](https://github.com/nknotts/nevermore-max-controller/blob/main/nevermoremax/__init__.py). [Here](https://github.com/nknotts/nevermore-max-controller/blob/287f9e5bbf49d54c297061889792f91caae9f16b/nevermoremax/__init__.py#L25) is where I register our module as a `bme280` and [here](https://github.com/nknotts/nevermore-max-controller/blob/287f9e5bbf49d54c297061889792f91caae9f16b/nevermoremax/__init__.py#L156) is where I send tvoc data. 

The resulting jsonrpc looks like:

Screen Shot 2022-06-25 at 6 33 52 AM

Once we get settled on the firmware logic, I plan on moving the logic within the firmware to the klipper module, at which point, I would write a sgp30 klipper module. If you would prefer to wait until that point, I would totally understand. I know its a bit weird to support something that doesn't exist yet. My only counterpoint is that the change is noninvasive and self-contained. I wanted to start the discussion with this pull request. It may be nice to offer a more generic solution to the problem of showing additional sensor information. I am not sure of the best way to go about this.

As far as testing goes, there are a few options

- If you have a `bme280`, you can added a `tvoc` entry [here](https://github.com/Klipper3d/klipper/blob/d3c4ba4839dd7a4339ae024752e6c6424884c185/klippy/extras/bme280.py#L465), or for a `htu21d` [here](https://github.com/Klipper3d/klipper/blob/d3c4ba4839dd7a4339ae024752e6c6424884c185/klippy/extras/htu21d.py#L239). Be sure to set it with random data or klipper will not detect any changes and wont send any updates. 
- Wait until there is a klipper sgp30 module
- My github repo contains a [sim](https://github.com/nknotts/nevermore-max-controller/blob/287f9e5bbf49d54c297061889792f91caae9f16b/sim.py)
  *  Create pseudo serial port pair with `socat -d -d pty,raw,echo=0,link=/dev/ttySIM0 pty,raw,echo=0,link=/dev/ttySIM1`
  * Run the sim, `python3 sim.py /dev/ttySIM0`
  * Configure the nevermore max module
```
[nevermoremax]
serial: /dev/ttySIM1

[temperature_sensor nevermore_intake]
sensor_type: nevermore_intake

[temperature_sensor nevermore_exhaust]
sensor_type: nevermore_exhaust
```

nknotts:
I have written a sgp30 Klipper module: https://github.com/nknotts/klipper/blob/sgp30/klippy/extras/sgp30.py. I will be submitting a pull request into mainline Klipper soon.

An example of the sgp30 module and the changes in this branch: 

Screen Shot 2022-07-29 at 1 50 37 PM

eduncan911:
@nknotts  @meteyou we now have an official Nevermore Discord server, where i'm starting talks about official software/firmware for these devices.  it would be great to get your inputs there as well.

https://discord.gg/invite/x6yggSpQgv

I know there's the hidden Voron sensors channel, that I don't have access to (because I don't have a serial) - as well as non-Voron users using Nevermores (Annex, Vcore, etc).  we created the discord server to be more open/public for those discussions and get non-Voron eyes on it as well.

meteyou:
Pls, create a new PR when this feature is implemented in klipper. I will close this PR to simply not have so many old PRs open in the list.

PR #949, opened 2022-07-08
Add linux/arm/v7 architecture to Docker builds

mkuf:
This PR adds the `linux/arm/v7` architecture to the docker Image Workflow to also support older SBCs. 

-Markus

### DIFF
--- a/.github/workflows/publish_docker.yml
+++ b/.github/workflows/publish_docker.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Build and push Docker image
         uses: docker/build-push-action@v2
         with:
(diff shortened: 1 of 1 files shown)


mkuf:
It seems, that npm install is failing because Cypress is incompatible with the new architecture and therefore no new images are pushed to the registry.

This can be mitigated by running the build stage only once on a supported platform and then copy the generated web files to the architecture specific runtime images.  
This is also how I currently build the mainsail images for prind, see https://github.com/mkuf/prind/blob/main/docker/mainsail/Dockerfile#L2

I'll test this in my fork and then open a new PR. 
You might want to revert this PR in the time being. 

-Markus

PR #950, opened 2022-07-08
ci: make cypress optional to resolve docker builds

pataar:
This'll omit cypress while building our Docker images.


pataar:
This also tests the building of Docker images on PR's. To prevent breaking the ci after merging.

PR #982, opened 2022-07-21
fix: hide TemperaturePanel if no sensors would be shown

ssendev:
If all Temperature sensors start with `_` an empty `TemperaturePanel` was shown.

Signed-off-by: Manuel Thomassen <thom@ssen.codes>

### DIFF
--- a/src/components/mixins/dashboard.ts
+++ b/src/components/mixins/dashboard.ts
@@ -20,8 +20,8 @@ export default class DashboardMixin extends BaseMixin {
         return this.$store.getters['printer/getAvailableHeaters'].length
     }
 
-    get printerAvailableSensorsCount() {
-        return this.$store.getters['printer/getAvailableSensors'].length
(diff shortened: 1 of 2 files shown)


meteyou:
Thx for the PR. Only one small addition. Pls also rename the function or add another getter if the same getter is used somewhere.

ssendev:
I changed the function name.

th33xitus:
Hi,

there is one spot in the code which also uses `printerAvailableSensorsCount` and wasn't replaced with the new variable name.
It is located here: https://github.com/ssendev/mainsail/blob/develop/src/pages/Dashboard.vue#L136

ssendev:
@th33xitus Strange I could have sworn I replaced all occurrences. Either way now it's done.

th33xitus:
LGTM!

th33xitus (approved):

PR #1002, opened 2022-08-04
fix(Heightmap): improve input validation for rename profile dialog

th33xitus:
Previously it was possible to:
- submit the rename dialog if the input field was empty
- rename a profile to a name that already exists
- rename a profile to 'default', the name 'default' is reserved by Klipper and will result in an error, making the profile to disappear

This PR will fix all the issues above by disabling the rename button in the rename dialog if one of the cases applies.

image
image
image
image


meteyou (approved):
LGTM

PR #1004, opened 2022-08-05
fix: divider in temperature presets is transparent

th33xitus:
### Solves the following issue:
For some reason the `v-divider` inside of the `v-menu` is rendered transparent and you can see the text underneath the `v-menu` through it.
22-08-05_20-57-18_chrome

Probably due to the case that it is set by Vuetify that way.
22-08-05_20-46-29_chrome
The expected behavior though would be, that the background color of the `v-menu` gets respected. Which, for some reason, is not the case. Presumably this might be a bug with Vuetify?

This PR solves the issue with applying a background color to that `v-divider` element as a workaround.
22-08-05_20-56-55_chrome

### DIFF
--- a/src/components/panels/TemperaturePanel.vue
+++ b/src/components/panels/TemperaturePanel.vue
@@ -31,7 +31,7 @@
                             </div>
                         </v-list-item>
                     </v-list>
-                    <v-divider></v-divider>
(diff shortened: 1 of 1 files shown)


meteyou (approved):

meteyou:
LGTM

PR #1018, opened 2022-08-10
feat: add warnings if gcodes/config root dirs don't exists

### LINKED ISSUE #984: Avoid displaying a wrong value for "Free Disk Space"
### Requested feature:

Display "unknown" if the free disk space cannot be determined instead of displaying 0.1kB.

When users see "0.1kB space free", they immediately panic and think their card is full. However, a misconfiguration of the gcode files path can lead to this, which is more likely.

### Additional information:

As discussed with meteyou on the Voron Discord

> Maybe it could be possible to check the root dir array. Pls write a FR on GitHub and I will check this.

meteyou:
this pr fix #984 


src/locales/en.json
```diff
@@ -175,6 +175,7 @@
         "Free": "Free",
         "FreeDisk": "Free disk",
         "GCodeFiles": "G-Code Files",
+        "GcodesRootDirectoryDoesntExists": "No Gcodes-Directory detected. Please check the path in your virtual_sdcard section in your Klipper config.",
```
th33xitus, line 178:
```suggestion
        "GcodesRootDirectoryDoesntExists": "No G-Code directory found. Please check option \"path\" in the [virtual_sdcard] section of the Klipper configuration.",
```

th33xitus (changes requested):

th33xitus (approved):

PR #1113, opened 2022-10-08
feat: add z_thermal_adjust to temperatures panel

### LINKED ISSUE #1105: Full temperature graph and panel implementation of [z_thermal_adjust] thermistor.
### Requested feature:

Moving from Alchemy engine to the new klipper implementation makes the [z_thermal_adjust] thermistor explicitly defined rather than a temperature_sensor reference.

Thus the [z_thermal_adjust] thermistor is no longer shown or controllable in the temp graph or panel. 

Its important to track the temp to see if frame temp is rising or falling between prints. If using VGB virtual gantry backers (possibly frame compensation) temps must be rising.

### Additional information:

https://www.klipper3d.org/Config_Reference.html#z_thermal_adjust
https://www.klipper3d.org/G-Codes.html#z_thermal_adjust

https://github.com/alchemyEngine/klipper_frame_expansion_comp/issues/7

meteyou:
fixes #1105 

thx @alchemyEngine for this Klipper feature!

### DIFF
--- a/src/components/panels/TemperaturePanel.vue
+++ b/src/components/panels/TemperaturePanel.vue
@@ -211,14 +211,14 @@
                                 @change="setVisible(dataset)"></v-checkbox>
                         </v-col>
                     </v-row>
-                    <v-row v-for="additionSensor in editHeater.additionalSensors" :key="additionSensor.name">
+                    <v-row v-for="additionalSensor in editHeater.additionalSensors" :key="additionalSensor.name">
                         <v-col class="col-12 py-1">
                             <v-checkbox
-                                v-model="additionSensor.bool"
-                                :label="$t('Panels.TemperaturePanel.ShowNameInList', { name: additionSensor.name })"
+                                v-model="additionalSensor.bool"
+                                :label="$t('Panels.TemperaturePanel.ShowNameInList', { name: additionalSensor.name })"
                                 hide-details
                                 class="mt-0"
-                                @change="setVisibleAdditionalSensor(additionSensor)"></v-checkbox>
+                                @change="setVisibleAdditionalSensor(additionalSensor)"></v-checkbox>
                         </v-col>
                     </v-row>
                     <v-row>
--- a/src/components/panels/TemperaturePanel.vue
+++ b/src/components/panels/TemperaturePanel.vue
@@ -149,7 +149,7 @@
                                                 {{ object.measured_min_temp }}°C
                                             </span>
                                         </v-tooltip>
-                                        <div v-for="(values, key) of object.additionSensors" :key="key">
+                                        <div v-for="(values, key) of object.additionalSensors" :key="key">
                                             <span v-if="values.bool" class="d-block">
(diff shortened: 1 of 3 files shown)


src/store/printer/getters.ts
```diff
@@ -281,6 +282,44 @@ export const getters: GetterTree<PrinterState, RootState> = {
...
+
+                if (value.temperature <= min_temp + split) icon = mdiThermometerLow
+                if (value.temperature >= max_temp - split) icon = mdiThermometerHigh
+
+                const additionalSensorBool = rootGetters['gui/getDatasetAdditionalSensorValue']({
+                    name: key,
+                    sensor: 'z_adjust',
+                })
+
+                const additionSensor: PrinterStateAdditionSensor = {
```
th33xitus, line 299:
```suggestion
                const additionalSensor: PrinterStateAdditionSensor = {
```

src/store/printer/getters.ts
```diff
@@ -281,6 +282,44 @@ export const getters: GetterTree<PrinterState, RootState> = {
...
+                    bool: additionalSensorBool,
+                    name: 'z_adjust',
+                    unit: 'μm',
+                    value: Math.round(value.current_z_adjust * 1000),
+                }
+
+                if (Math.abs(value.current_z_adjust) >= 0.1) {
+                    additionSensor.value = Math.round(value.current_z_adjust * 1000) / 1000
```
th33xitus, line 307:
```suggestion
                    additionalSensor.value = Math.round(value.current_z_adjust * 1000) / 1000
```

src/store/printer/getters.ts
```diff
@@ -281,6 +282,44 @@ export const getters: GetterTree<PrinterState, RootState> = {
...
+                    additionSensor.unit = 'mm'
```
th33xitus, line 308:
```suggestion
                    additionalSensor.unit = 'mm'
```

src/components/panels/TemperaturePanel.vue
```diff
@@ -243,7 +243,7 @@ import { Mixins } from 'vue-property-decorator'
 import { capitalize, convertName } from '@/plugins/helpers'
 import { Debounce } from 'vue-debounce-decorator'
 import { GuiPresetsStatePreset } from '@/store/gui/presets/types'
-import { PrinterStateTemperatureObject } from '@/store/printer/types'
+import { PrinterStateAdditionSensor, PrinterStateTemperatureObject } from '@/store/printer/types'
```
th33xitus, line 246: Honestly, i would prefer renaming the type to `PrinterStateAdditionalSensor`, so that we have a matching nomenclature.
I quite dislike the mixed use of `addition` and `additional`, unless there is a specific reason for it i might miss right now?
```suggestion
import { PrinterStateAdditionalSensor, PrinterStateTemperatureObject } from '@/store/printer/types'
```

th33xitus (changes requested):

th33xitus (approved):

Issue #897, opened 2022-06-12
Power On button when printer is disconnected.


LazaroFilm:
### Requested feature:

When the printer is off, I need to go into the top right power menu to turn it on, I would love a Power on button in the main view when the printer the Printer switch is detected in the OFF position.
![Screen Shot 2022-06-12 at 1 47 22 PM](https://user-images.githubusercontent.com/18351100/173246711-38a126c3-3d6f-4281-927d-964e7571dc6e.png)

### Solves the following problem:

Adding a Power On Printer button in the main dashboard area when the Printer switch is Off.

meteyou:
Something like this is on our to-do list, but I think I will only display one panel with "print is off -> switch on" or something similar.

LazaroFilm:
Amazing news!  .... waits... (jk. You guys are doing an amazing job already, no rush, take your time, I'll just be here, waiting patiently)

LazaroFilm:
> > ### Requested feature:
> 
> > When the printer is off, I need to go into the top right power menu to turn it on, I would love a Power on button in the main view when the printer the Printer switch is detected in the OFF position. ![Screen Shot 2022-06-12 at 1 47 22 PM](https://user-images.githubusercontent.com/18351100/173246711-38a126c3-3d6f-4281-927d-964e7571dc6e.png)
> 
> > 
> 
> > ### Solves the following problem:
> 
> > Adding a Power On Printer button in the main dashboard area when the Printer switch is Off.
> 
> > 
> 
> > ### Additional information:
> 
> > _No response_
> 
> 
> 
> Hi, how to make Power on button ?

The image is a "photoshop" of what it could look like. To my understanding, as of now, it is not possible to have a power on button.

jordandrako:
Why wouldn't it be possible? Wouldn't it just be a matter of having an option to expose specific power devices on the dashboard (or even the top bar would be nice)

PR #1118, opened 2022-10-10
refactor: rework of the KlippyState panel

th33xitus:
Often there is quite a bit confusion about error messages. Users often think, those messages come from Mainsail, although they are either directly coming from Klipper or Moonraker. This PR has the goal to make it a bit more clear where the errors come from. For that, the Klippy State Panel was reworked quite drastically. Depending on the status, different colors were choosen. Additionally, it is now possible to directly download the Klipper and/or Moonraker logfiles from that panel.
A simple title should make it more clear, which component is reporting the corresponding message and in turn make it more clear for the user.

## Before:
image
image
image
image

## After:
image
image
image
image

### DIFF
--- a/src/components/panels/KlippyStatePanel.vue
+++ b/src/components/panels/KlippyStatePanel.vue
@@ -1,10 +1,10 @@
 <template>
     <div v-if="klipperState !== 'ready' && socketIsConnected">
         <v-container v-if="klippyIsConnected" class="pa-0 pb-6">
-            <v-alert :color="messageType" dense text border="left" class="mb-0">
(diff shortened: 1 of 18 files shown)


meteyou (approved):

LazaroFilm:
I still wished there was an option to add a [POWER PRINTER ON] button in this panel when the MCU is disconnected and the Printer switch is OFF. It would be nice to be welcomed with a on button instead of a RED ERROR when starting the printer...
https://github.com/mainsail-crew/mainsail/issues/897

meteyou:
@LazaroFilm something like this is on our Todo list...

PR #1134, opened 2022-10-30
refactor(KlippyStatePanel): display buttons as outlined text buttons

th33xitus:
### Before:
22-10-10_22-41-58_chrome

### After:
image


meteyou (approved):

LazaroFilm:
I still wished there was an option to add a [POWER PRINTER ON] button in this panel when the MCU is disconnected and the Printer switch is OFF. It would be nice to be welcomed with a on button instead of a RED ERROR when starting the printer...
https://github.com/mainsail-crew/mainsail/issues/897

PR #1192, opened 2022-12-13
feat: add table view for print status stats

### LINKED ISSUE #891: Other views in history tab
### Requested feature:

At this moment in a history tab you can view statistics only in "circle view" but it isn't the best in all case. So I suggest adding something like "table view". So if someone wants to view all statistics he can use table view. 

obraz_2022-06-12_130053709
It will solve in some cases the problem of names stacking like this.

### Additional information:

_No response_

meteyou:
this PR will fix #891 

Chart view:
image

Table view:
image

### DIFF
--- a/src/components/charts/HistoryAllPrintStatusTable.vue
+++ b/src/components/charts/HistoryAllPrintStatusTable.vue
@@ -1,0 +1,47 @@
+<template>
+    <v-simple-table>
+        <tbody>
+            <tr v-for="status in printStatusArray" :key="status.name">
+                <td>{{ status.displayName }}</td>
+                <td class="text-right">{{ status.value }}</td>
+            </tr>
+        </tbody>
+    </v-simple-table>
+</template>
+
+<script lang="ts">
+import Component from 'vue-class-component'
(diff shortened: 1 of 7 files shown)


src/components/charts/HistoryAllPrintStatusTable.vue
```diff
@@ -0,0 +1,47 @@
...
+                <td>{{ status.displayName }}</td>
+                <td class="text-right">{{ status.value }}</td>
+            </tr>
+        </tbody>
+    </v-simple-table>
+</template>
+
+<script lang="ts">
+import Component from 'vue-class-component'
+import { Mixins, Watch } from 'vue-property-decorator'
```
th33xitus, line 14: unused import

th33xitus (changes requested):

src/components/charts/HistoryAllPrintStatusTable.vue
meteyou: Fixed with the new commit

th33xitus (approved):

PR #1215, opened 2023-01-05
locale(ko): update Korean localization

ZISQO:
fixing korean language pack : ) thanks!


th33xitus:
Translation incomplete. Lots of keys missing. @ZISQO are you able to update those?

th33xitus (changes requested):
Translation incomplete. Lots of keys missing. @ZISQO are you able to update those?

PR #1228, opened 2023-01-15
Mainsail for belt printer

milkapot:
Mainsail version for belt printer with Y axis probe, still need to be tested

### DIFF
--- a/src/components/dialogs/TheManualProbeDialog.vue
+++ b/src/components/dialogs/TheManualProbeDialog.vue
@@ -15,11 +15,11 @@
             <v-container>
                 <v-row>
                     <v-col class="d-flex align-center justify-center">
-                        <span class="text-h5">{{ z_position_lower }}</span>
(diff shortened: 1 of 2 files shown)


meteyou:
Please don't overwrite the default function. It would be nice to have a setting to enable this feature in the interface settings to switch between Z and Y "baby-stepping".

milkapot:
Hello, 
Sorry that was the only way that i could use Y-offset for my probe on my belt printer. How can we implement this feature option? What can I do to help for this feature. I know that Mainsail is compiled but I don't know how to do it with a new branch.
Please let me know what can I do to help and support you on this feature.
have a good day.

meteyou:
Please add the option to switch between Z and Y via interface settings. Then I think you have to hide the safe_z_offset buttons because this klipper gcode only works for Z. I cannot merge an option that deletes the Z option because this is the most use case.

meteyou:
I am closing this PR because there is no activity

PR #1316, opened 2023-03-20
feat: add facility to Scan Metadata from G-code Files

siparker:
When viewing a mounted network drive gcode files that are added to the network share and not uploaded via Mainsail do not contain any meta data or thumbnails. 
This was originally raised as an issue within Moonraker https://github.com/Arksine/moonraker/issues/614 and determined to be an issue with some mount options, for example cifs, did not properly run inotify events which moonraker relies on. 
There was already a moonraker API endpoint /server/files/metascan that can scan existing files so we just added the option to right click and scan meta data via Mainsail instead of trying to cover every mount type and inotify compatibility issues.

### DIFF
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -1385,35 +1385,9 @@ export default class GcodefilesPanel extends Mixins(BaseMixin, ControlMixin) {
     }
 
     scanMeta(item: FileStateFile) {
-        const filepath = this.currentPath + '/' + item.filename
-        const file = filepath.slice(0, 1) === '/' ? filepath.slice(1) : filepath
-        const data = {
-            filename: file,
-        }
-        item.small_thumbnail = undefined
-        axios
-            .post(this.apiUrl + '/server/files/metascan', data, {
-                headers: { 'Content-Type': 'multipart/form-data' },
-            })
-            .then((response) => {
-                const $this = this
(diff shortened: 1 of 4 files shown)


meteyou:
Thx for your PR. I didn't know that Moonraker had this function by now.

Why do you use the HTTP request, not the WebSocket, to update the metadata?

siparker:
No specific reason other than that is what i had been testing with using postman so i knew the behaviour. The api endpoint has some slightly odd behaviours. Such as if the file doesnt exist then it automatically scans the entire document tree.
If you prefer i could redo it using websockets?

meteyou:
i would prefer to use websocket and send the result to the action files/getMetadata to store the new metadata in the state

meteyou:
@siparker pls enable pushing to this branch/pr. I refactored this to switch to websocket requests, but i'm not allowed to push it in your branch.

siparker:
> @siparker pls enable pushing to this branch/pr. I refactored this to switch to websocket requests, but i'm not allowed to push it in your branch.

Hi, After a few mins of searching apparently there is a bug to allow edits for maintainers in a push request from an Org. https://github.com/orgs/community/discussions/5634 

I have added you as a collaborator on the develop branch, hopefully that allows it?

meteyou:
Ok. but there was also an issue on my side. I also had to use a own taken for this branch. This could also be my issue.

But now it worked. I pushed my changes. Please, double-check it!

meteyou (approved):

Issue #1317, opened 2023-03-21
Add support for webrtc streaming for IP Cameras


slynn1324:
### Requested feature:

I use a wyze cam, and run a wyze-bridge (https://github.com/mrlt8/docker-wyze-bridge) docker container on my network that publishes a webrtc stream that works great (low latency, reliable).  I would like to be able to integrate the remote webrtc video stream as a WebCam in Mainsail.  

The existing "web-rtc (camera-streamer)" webcam option seems to have a lot of code specifically pointing to Raspberry Pi functionality.  From searching some documentation, it appeared that the "IP Camera" option is intended for use to have the browser directly reach out to a different address - and not tunnel any data through the Mainsail host.  

I have a working PR that I will open shortly.

PR #1318, opened 2023-03-21
feat: add WebRTC (MediaMTX / rtsp-simple-server) webcam mode

slynn1324:
Adds support for ws://... urls in the IP Camera webcam type.

See https://github.com/mainsail-crew/mainsail/issues/1317.


meteyou:
Thx for you your PR! The main problem with WebRTC is that there are so many different "handshakes" for WebRTC. It's not possible to have one webrtc mode to support all streamers.

I only would prefer to have a separate mode for this webcam style. It looks like the "wyze-bridge" is using the rtsp-simple-server in the background for the webrtc stream. So I think the name "WebRTC (rtsp-simple-server)" would be a good name for this mode.

The mode "IP-Webcam" is for generic IP-Webcams.

slynn1324:
Fair enough... I didn't dig into all the details of the underlying protocols (just hacked together what worked from an example) but that makes sense.  I'll look at refactoring this out to a separate camera type and update this PR.  Thanks!

slynn1324:
Updated as a new camera type.  I think I poked around and found the necessary references that the camera types need expanded.  Please take a look.

src/components/webcams/WebrtcRTSPSimpleServer.vue
```diff
@@ -0,0 +1,251 @@
...
+
+        this.webRtcWs.onmessage = (msg: any) => this.webRtcOnIceServers(msg)
+    }
+
+    webRtcTerminate() {
+        this.webRtcTerminated = true
+
+        if (this.webRtcWs) {
+            try {
+                this.webRtcWs.close()
```
meteyou, line 131: this.webRtcWs?.close() should be enough here. i dont think you need a try...catch statement

src/components/webcams/WebrtcRTSPSimpleServer.vue
```diff
@@ -0,0 +1,251 @@
...
+            } catch (err) {
+                // ignore
+            }
+        }
+        if (this.webRtcPc) {
+            try {
+                this.webRtcPc.close()
```
meteyou, line 138: this.webRtcPs?.close() should be enough here. i don't think you need a try...catch statement

src/components/webcams/WebrtcRTSPSimpleServer.vue
```diff
@@ -0,0 +1,251 @@
...
+            this.webRtcPc.close()
+            this.webRtcPc = null
+        }
+
+        if (this.webRtcTerminated) {
+            return
+        }
+
+        this.webRtcRestartTimeout = window.setTimeout(() => {
+            this.webRtcRestartTimeout = null
```
meteyou, line 246: pls clear the timeout in the webRtcStart function and set it there to null

src/components/webcams/WebrtcRTSPSimpleServer.vue
```diff
@@ -0,0 +1,251 @@
...
+        try {
+            this.video.pause()
+        } catch (err) {
+            // ignore - more important to shut down the sockets.
+        }
+
+        this.webRtcTerminate()
+    }
+
+    visibilityChanged(isVisible: boolean) {
```
meteyou, line 75: i would not use the visibilty check for webRTC, because webRTC needs too long to reconnect and it also need only a small bandwidth to work.

src/components/webcams/WebrtcRTSPSimpleServer.vue
```diff
@@ -0,0 +1,251 @@
+<style scoped>
```
meteyou, line 1: pls move the style tag to the end of the vue file (i'm refactoring every file, when i see this "wrong" sort)

meteyou (commented):
in general, this already looks very good. i added a few comments, but after that it should fit. i will try to install this streamer to be able to test this.

slynn1324:
See how that is.  I reworked the start/stop logic to track when the panel is actively collapsed vs. just scrolled out of view.  I also had to make a small change to TheSettingsMenu.vue so that the component would be destroyed if you click on the 'x' to close to the settings dialog without either cancelling or updating the camera in the dialog first.  I think this strikes a good balance between not reconnecting too frequently yet only keeping the streaming connection open if it stands to be displayed.   

I think I addressed all of the other comments.  

Thanks.

meteyou:
I cleaned the file and added TS types as well as possible. Yesterday also, the project rtsp-simple-server was renamed to MediaMTX. So I also renamed it in this PR. I tested it with a newly installed MediaMTX with a raspberry webcam. Please double-check if it also works in your case, and then this PR will be merge-ready.

slynn1324:
Apologies for the delay in getting back here.  I pulled the updates and installed them and it seems to work fine with my setup here and the wyze-bridge stream.  

I did notice 2 things however that were working better (in my opinion) prior to your updates:

1) The web socket connection would not be opened when the panel was not expanded.  The way that the panels seem to work on the dashboard (presumably due to animations), the components inside are mounted even when they are not visible.  I do not think that the additional load on the backing servers (or the bandwidth) should be used to open the stream if it is not visible.

2) The controls attribute was removed from the <video> element.  The controls allow to stop and start the stream - again for bandwidth purposes if desired.  What is even more useful is on mobile (at least iOS devices), including the controls allows for opening the video in full screen natively.  While not having any controls is probably a bit more consistent with some of the other video formats - I'd encourage to include them here where it's natively supported.

I typically have mainsail up on 3 devices simultaneously:  A monitor by the printer -- no camera needed there. I keep the Webcam panel minimized), my main PC in my office where I do most of my work, but monitor the printer from another room in my house, and then checking in from iPhone or iPad devices when I'm not in one of those places.  

Of the two issues, I think #1 should be looked at more closely and corrected for to be kind to users bandwidth and device needs, and #2 is more of a preference.

meteyou:
@slynn1324 thx for your high-level feedback! You are right about point 1. I saw it as "unimportant" when refactoring and wanted to remove unnecessary complexity, but with your explanation of your workflow, this sounds very reasonable! Do you want to add this back, or should I do it?

I intentionally removed the controls because not all functions work and would only lead to "wrong bug reports". I did the same with the WebRTC (camera-streamer) client. But I wouldn't know how to hide specific controls, so I removed them all. We can gladly enable them again if you know how to show and hide them to ensure 100% compatibility.

slynn1324:
If you want to add back #1 as you'd like it to be that'd be great.  It might be a bit before I'm able to get back into the code.  

As for the controls - I would doubt there is a way to use only some of the native controls.  On my iOS devices, it shows up with the native player controls which then allows full screen and pausing etc.  It also displays that it's a live stream.  Would this be something to consider an extra option in the webcam configuration panel about whether the native controls should be displayed or not?  I could really see different users wanting it different ways... depending on their preferences.

meteyou:
@slynn1324 i pushed a few changes to fix the closed panel behavior. it was included in the code, but with a bug.

Please test the build again.

slynn1324:
Tested here and seems to be working fine.  I would think this is good to merge whenever you're ready.

src/components/panels/WebcamPanel.vue
```diff
@@ -63,6 +63,9 @@
                     <template v-else-if="currentCam.service === 'webrtc-camerastreamer'">
                         <webcam-webrtc-camerastreamer :cam-settings="currentCam" />
                     </template>
+                    <template v-else-if="currentCam.service === 'webrtc-mediamtx'">
+                        <webcam-webrtc-mediamtx :cam-settings="currentCam" :collapsible="$route.fullPath !== '/cam'" />
```
meteyou, line 67:
```suggestion
                        <webcam-webrtc-mediamtx :cam-settings="currentCam" />
```

this variable is not used in the component anymore

src/components/webcams/WebrtcMediaMTX.vue
```diff
@@ -0,0 +1,238 @@
...
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop, Ref, Watch } from 'vue-property-decorator'
+import BaseMixin from '@/components/mixins/base'
+
+@Component
+export default class WebrtcRTSPSimpleServer extends Mixins(BaseMixin) {
+    @Prop({ required: true }) camSettings: any
+    @Prop({ default: false }) collapsible: boolean | undefined
```
meteyou, line 27:
```suggestion

```

this prop is not used in this component anymore

src/components/TheSettingsMenu.vue
```diff
@@ -55,7 +55,10 @@
                             ref="settingsScroll"
                             :class="'settings-tabs ' + (isMobile ? '' : 'height500')"
                             :options="{ overflowBehavior: { x: 'hidden' } }">
-                            <component :is="'settings-' + activeTab + '-tab'" @scrollToTop="scrollToTop"></component>
+                            <component
+                                :is="'settings-' + activeTab + '-tab'"
+                                v-if="showSettings"
```
meteyou, line 60: Why do you add here `v-if="showSettings"`?

meteyou:
Ok. I reviewed the code in detail and have found some little refactorings and one question. But the rest looks good to me. We also have to update this branch before merging it because other webcam clients were added to the developer branch, and there are merge conflicts right now.

meteyou:
@slynn1324 any updates here?

meteyou (approved):

meteyou:
@slynn1324 can you share any information how to setup this webcam in wyze-bridge and which url you need to use in mainsail?

Naninani:
@meteyou @slynn1324 Could you share instructions on how to setup?

PR #1327, opened 2023-03-28
locale(cz): add Czech localization

PavelTajdus:
Hello,
I created localization file and tested on local enviroment. I was fine tuning some of czech sentences, and I think its ready to release to public :-)

Thank you for great work on Mainsail.


PavelTajdus (commented):
Code formated with Prettier.

th33xitus (approved):

PR #1336, opened 2023-04-10
feat: customize sidebar navi

meteyou:
With this PR, you can add custom links or sort/hide nav points (in the interface settings).

To add new nav points, you have to add a JSON file in `config/.theme/navi.json` with the following example config:

```json
[
  {
    "title":"Custom1",
    "href":"https://www.google.com"
  },
  {
    "title":"Moonraker",
    "href":"/server/info",
    "target": "_blank",
    "position": 90,
    "icon": "M5 5H7V11H5V5M10 5H8V11H10V5M5 19H7V13H5V19M10 13H8V19H10V17H15V15H10V13M2 21H4V3H2V21M20 3V7H13V5H11V11H13V9H20V15H18V13H16V19H18V17H20V21H22V3H20Z"
  }
]
```
per default, all nav points have 10 steps between each other positions. That means:
- dashboard => 10
- console => 20
- heightmap => 30

Thus it should be possible to integrate each custom nav point on each position by choosing one position value between them. You can also use custom icons. You only have to use the JS icon content of MDI (https://materialdesignicons.com/).


th33xitus (approved):

jeno007:
I'd like to use this feature to add Obico's web UI to mainsail and the mainsail UI to OrcaSlicer's device tab.
The problem I'm facing is that loading Obico's web as _self overwrites the whole page making the sidebar unavailable. As Orca does not support tabs on their device tab, after opening Obico, there is no way back to the rest of mainsail. Specifying Obico as _blank does not help either because tabs are not supported so does not open. I'think it would help to open Obico inside a properly setup div to keep the sidebar working. Unfortunately I'm not familiar with typescript to make it working.

PR #1341, opened 2023-04-17
fix: add gcode offset to live position in gcodeviewer

meteyou:


zellneralex (approved):
Tested on the V400 with a large negativ gcode offset. Looks good now.

image

th33xitus (approved):

Issue #1163, opened 2022-11-21
Add nginx reverse proxy support


EricZimmerman:
### Requested feature:

I tried for hours to get this working with nginx to no avail. Most apps seem to want a urlprefix for things to work and this was my issue I believe since the app always looked at the root of the url for various files. 

I tried rewrite rules and a bunch of other things. Other examples are sonarr and radar, among others 

### Solves the following problem:

Allow for security to be managed in nginx including SSL, authentication and 100 other things.

meteyou:
> Allow for security to be managed in nginx including SSL, authentication and 100 other things.
All of that is possible with the current state.

EricZimmerman:
so, is this something where you would just replicate the existing docs on nginx on another nginx instance? i was trying to essentially proxy the nginx instance on the pi to my external nginx box.

EricZimmerman:
Except for webcam access, or at least it didn't when I did the subdomain
approach.

One, or both, options should work ideally

SoloTSi97:
@EricZimmerman Obviously, this issue is still open but I wonder if you made any further progress on your configuration? This is exactly my setup and I'm having similar struggles. Thanks!

EricZimmerman:
Nope. I gave up. Not worth the hassle without basic support for this imo. 

I just use zerotier

Roguyt:
I currently have a working solution, which the only caviat is regarding the basic authentification. Sometimes it doesn't renew so I need to open like the webcam stream URL to prompt it again and I'm good to go.

![image](https://user-images.githubusercontent.com/8149856/231693191-8d4e8e6f-3d7c-405b-80b6-da6355a750e8.png)

EricZimmerman:
Can you post it?

Roguyt:
Sure thing.

All its missing is Mainsail having native authentication to remove the basic nginx authentication.

```nginx
server {
        listen 80;
        server_name mainsail.baguette.fr;

        location ~ /\.well-known/acme-challenge {
                root /var/www/html;
                allow all;
        }

        location / {
                return 301 https://$server_name$request_uri;
        }
}

server {
        listen 443 ssl;
        server_name mainsail.baguette.fr;

        add_header Strict-Transport-Security "max-age=31536000";
        ssl_certificate /etc/letsencrypt/live/mainsail.baguette.fr/fullchain.pem;
        ssl_certificate_key /etc/letsencrypt/live/mainsail.baguette.fr/privkey.pem;
[...]

PR #1359, opened 2023-04-29
feat: add configurable base URL

your-friend-alice:
This PR will fix https://github.com/mainsail-crew/mainsail/issues/1163

Adds a `VITE_BASE` environment variable at build-time that allows the user to serve the app from non-root URL paths.

Signed-off-by: Alice <alice@dn3s.me>

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+VITE_BASE=/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,20 @@ jobs:
       - name: Copy output
         run: cp ${{ github.workspace }}/dist/mainsail.zip mainsail-latest.zip
 
+      - name: Build With Sub-Path
+        run: npm run build
+        env:
+          VITE_BASE: /mainsail/
+
+      - name: Copy output
+        run: cp ${{ github.workspace }}/dist/mainsail.zip mainsail-subpath-latest.zip
+
       - uses: actions/upload-artifact@v2.2.4
         with:
           name: mainsail-latest.zip
           path: mainsail-latest.zip
+
+      - uses: actions/upload-artifact@v2.2.4
+        with:
+          name: mainsail-subpath-latest.zip
+          path: mainsail-subpath-latest.zip
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,20 +39,7 @@ jobs:
       - name: Copy output
         run: cp ${{ github.workspace }}/dist/mainsail.zip mainsail-latest.zip
 
-      - name: Build With Sub-Path
-        run: npm run build
-        env:
-          VITE_BASE: /mainsail/
-
-      - name: Copy output
-        run: cp ${{ github.workspace }}/dist/mainsail.zip mainsail-subpath-latest.zip
-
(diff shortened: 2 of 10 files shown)


your-friend-alice:
this is now working on my machine, so it should be good to review. the only thing not checked is the github actions changes (also, it's up to the maintainers whether you want to have a second build or not, it would impact your project's storage usage slightly)

your-friend-alice:
note: this addresses all the issues in my own setup, but there's probably other paths I've missed. I'm not a web developer so this was all pulled together through trial and error, there should probably be a test case added to catch any regressions / missed cases

meteyou:
Thx for your PR. This PR with the changed build workflow would "kill" all default installations. So the default should not be changed.

I have to check the rest on my test rig.

meteyou:
I would prefer a dynamic solution (editable via config.json file). I see a separate build for subdirectories as a workaround. I couldn't test the build yet, but I wouldn't be sure now if the API would also run via reverse proxy.

your-friend-alice:
that would be ideal, unfortunately a lot of the static assets are compiled at build time that contain absolute paths, it's possible some of them could be changed to relative paths but with the virtual multi-page system I'm not sure how that would interact. I don't love this solution either but I didn't see any other approach to take that wouldn't involve major architectural changes.

If you don't want to publish multiple builds, I can remove the github actions changes, I think just upstreaming this build option would already be a huge win for weird nerds like myself who want to serve each webapp at its own subpath. I'd be happy to write some docs for this feature as well, let me know where to look and if that's in-repo I can put it in this PR.

As for moonraker, this PR shouldn't change anything about where it looks for moonraker (`/moonraker`), I'd like to make that configurable as well but that's outside of the scope of this PR.

meteyou:
Are you able to use the Moonraker update manager to use the subpath?

your-friend-alice:
i don't think so, looks like that would require adding functionality to Moonraker to set build-time environment variables

meteyou:
nono... moonraker is not building mainsail. it just download the prebuild zip from the latest release.

your-friend-alice:
looks like right now it's hardcoded to just use the first asset associated with a Release:
https://github.com/Arksine/moonraker/blob/fd5ea0c6a45fb04baf21014d1505f3d4ade791db/moonraker/components/update_manager/update_manager.py#L1277-L1278

So unless that's updated to iterate through the release assets with some name matching logic exposed as a config option, I should remove the github actions part from this PR.

I'm busy with work for the next while but I can try to follow up on this later this week, either creating a moonraker PR to add that logic to its updater or just removing the release part of this PR if you'd rather defer that for now

your-friend-alice:
Sorry for the delay. I've reverted the commit that adds the github builds, that should prevent any issues with Moonraker's auto-updater accidentally grabbing the wrong build. I've also rebased to clear up some merge commits, and it looks like https://github.com/mainsail-crew/mainsail/pull/1336 made a lot of this PR's changes redundant, so the diff should look a lot smaller now

meteyou:
i will close this PR, because of no response... Please open a new PR, if you have an idea how to fix this issue.

your-friend-alice:
totally understandable, I haven't had time to work on this, and it's sat too long by now

PR #1369, opened 2023-05-02
feat: Add additional Timelapse settings

mikebcbc:
## Description

Adds additional settings for rotation, flip X, and flip Y to the timelapse settings tab.

[I made a PR on moonraker-timelapse](https://github.com/mainsail-crew/moonraker-timelapse/pull/112) to update their configuration documents to accompany these changes. This is tested and working, however it's not documented in their configuration documents - only in their source.

image

Nope. Should be good to go.


meteyou:
Time lapse uses the webcam settings. So this is not necessary.

meteyou:
i don't see any advantage in only rotating the timelapse recording. please also see the webcam application that everything is displayed correctly.

mikebcbc:
> i don't see any advantage in only rotating the timelapse recording. please also see the webcam application that everything is displayed correctly.

Gotcha. I can see your point. Are you cool if i update this PR to make the `rotation` setting available in the webcam settings in the UI for mainsail? It's not currently available, and if we don't want time lapse specific settings, I can at least make the rotation setting available for the global webcam settings. That would solve my issue as well. Moonraker supports it and Mainsail doesn't have an option for it right now.

meteyou:
Yes. This would fit. Some webcam modes have it already. Which one do you use?

mikebcbc:
> Yes. This would fit. Some webcam modes have it already. Which one do you use?

I use UV4L-MJPEG. It's not default available for that one, looks like only for adaptive mjpeg. I can add it for UV4L, rotation works when placed in my moonraker-timelapse config.

Issue #1322, opened 2023-03-24
Printer name in browser tab during printing


ckvsoft:
### Requested feature:

It would be nice if the name of the printer remained in the tab when printing or was appended to the progress display in the tab.
With several printers this is annoying.

mikebcbc:
I made a PR that got merged that adds this. Once it's released i think this issue can be closed

meteyou:
normally i close the FR when i merge (automatically by github because the issue is linked in the PR)

mikebcbc:
Cool!

PR #1371, opened 2023-05-03
feat: Add printer name to browser tab while printing or complete

mikebcbc:
Adds the printer name to the browser tab during printing and complete statuses.

## Description

This PR addresses #1322 and adds the printer name to the browser tab while the printer is printing, or when the print is complete.

#1322 

image

No post-deployment tasks.

### DIFF
--- a/src/components/settings/SettingsUiSettingsTab.vue
+++ b/src/components/settings/SettingsUiSettingsTab.vue
@@ -341,7 +341,7 @@ export default class SettingsUiSettingsTab extends Mixins(BaseMixin) {
     }
 
     get disableFanAnimation() {
-        return this.$store.state.gui.uiSettings.hideSaveConfigForBedMash ?? false
+        return this.$store.state.gui.uiSettings.disableFanAnimation ?? false
     }
 
     set disableFanAnimation(newVal) {
--- a/src/components/settings/SettingsUiSettingsTab.vue
+++ b/src/components/settings/SettingsUiSettingsTab.vue
@@ -341,7 +341,7 @@ export default class SettingsUiSettingsTab extends Mixins(BaseMixin) {
     }
 
     get disableFanAnimation() {
-        return this.$store.state.gui.uiSettings.disableFanAnimation ?? false
+        return this.$store.state.gui.uiSettings.hideSaveConfigForBedMash ?? false
     }
 
     set disableFanAnimation(newVal) {
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -54,7 +54,7 @@
             "Pause": "Pause Print",
             "PrinterOff": "Printer OFF",
             "Printing": "{percent}% Printing - {filename} {name}",
-            "PrintingETA": "Complete - {filename} {name}"
+            "PrintingETA": "{percent}% Printing - ETA: {eta} - {filename} {name}"
         },
         "TopBar": {
             "CannotUploadTheFile": "Cannot upload the file!",
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -49,12 +49,12 @@
             "TitleUnderVoltageDetected": "Under-Voltage Detected"
         },
         "Titles": {
-            "Complete": "Complete - {filename} - {name}",
+            "Complete": "Complete - {filename} {name}",
             "Error": "ERROR",
             "Pause": "Pause Print",
             "PrinterOff": "Printer OFF",
-            "Printing": "{percent}% Printing - {filename} - {name}",
-            "PrintingETA": "{percent}% Printing - ETA: {eta} - {filename} - {name}"
+            "Printing": "{percent}% Printing - {filename} {name}",
+            "PrintingETA": "Complete - {filename} {name}"
         },
         "TopBar": {
             "CannotUploadTheFile": "Cannot upload the file!",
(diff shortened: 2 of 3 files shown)


src/components/settings/SettingsUiSettingsTab.vue
```diff
@@ -341,7 +341,7 @@ export default class SettingsUiSettingsTab extends Mixins(BaseMixin) {
     }
 
     get disableFanAnimation() {
-        return this.$store.state.gui.uiSettings.hideSaveConfigForBedMash ?? false
+        return this.$store.state.gui.uiSettings.disableFanAnimation ?? false
```
meteyou, line 344: please create a separate PR for this bugfix

src/locales/en.json
```diff
@@ -49,12 +49,12 @@
             "TitleUnderVoltageDetected": "Under-Voltage Detected"
         },
         "Titles": {
-            "Complete": "Complete - {filename}",
+            "Complete": "Complete - {filename} - {name}",
```
meteyou, line 52: i would recommend to change the name of these translations, that all translators see a new key to translate

meteyou (changes requested):

meteyou:
Thx for your PR. I have only one question about it. What happens when the user don't set a printername in the settings?

src/locales/en.json
mikebcbc: Can you give an example of what you mean? Do you want me to go through every translation and update? Or change the key name? Won't that break the other translations if I do the second one?

src/components/settings/SettingsUiSettingsTab.vue
mikebcbc: done: https://github.com/mainsail-crew/mainsail/pull/1381

mikebcbc:
> Thx for your PR. I have only one question about it. What happens when the user don't set a printername in the settings?

@meteyou - good question, I updated this PR to handle that. Before it would just show the hyphen, but now I am conditionally showing the hyphen. Take a look at the PR. Here's two examples:

printername defined:
image

printername not defined:

image

src/locales/en.json
meteyou: That's what changing the key is precisely about. Just as an example:

You only change the English translation here. The key remains the same, so no translator is made aware that there are new translations that are not translated. If we take DE as an example, without this change, the old text is still displayed without the name.

So it is better to change the key, which does not exist as a translation, but it has a correct fallback in English with all variables or texts it should have.

An working "new" key would be "CompleteWithPrintername"...

meteyou:
hmm... maybe my "idea" with adding the `{name}` in the translation itself was wrong for this use case. maybe you should just use getter to add " - <name>", when it exists and we also don't have an issue with the "new" translation keys.

src/locales/en.json
mikebcbc: Ahh, understood! That makes sense - I went ahead and updated this! Thanks for clarifying!

mikebcbc:
> hmm... maybe my "idea" with adding the `{name}` in the translation itself was wrong for this use case. maybe you should just use getter to add " - ", when it exists and we also don't have an issue with the "new" translation keys.

@meteyou Are you saying to force it in the getter? Like this:

```vue
return i18n.t('App.Titles.PrintingETA', {
                    percent: percent,
                    filename: state.printer?.print_stats?.filename,
                    eta
                }) + state.gui?.general.printername ? `- ${state.gui?.general.printername}` : ""
``` 

If so, I feel like that's not very clean... I feel like it would be cleaner to just supply the `- {name}` and update the translation keys. If we don't want to worry about translation, since it's all the same no matter the language I can just update every translation and append this where it fits... what are you thoughts on that?

meteyou:
i would use write it a little bit different (just for readability):
```ts
let output = i18n.t('App.Titles.PrintingETA', {
	                    percent: percent,
	                    filename: state.printer?.print_stats?.filename,
	                    eta
	                })

// add printer name if it exists
if (state.gui?.general.printername) output += ` - ${state.gui.general.printername}`

return output
```

I just don't like to add "decorators" in a translation variable.

just my 2 cents about this topic...
(my solution have not to be the best way... its just my style to do it...)

mikebcbc:
> 

@

> i would use write it a little bit different (just for readability):
> 
> ```ts
> let output = i18n.t('App.Titles.PrintingETA', {
> 	                    percent: percent,
> 	                    filename: state.printer?.print_stats?.filename,
> 	                    eta
> 	                })
> 
> // add printer name if it exists
> if (state.gui?.general.printername) output += ` - ${state.gui.general.printername}`
> 
> return output
> ```
> 
> I just don't like to add "decorators" in a translation variable.
> 
> just my 2 cents about this topic... (my solution have not to be the best way... its just my style to do it...)

@meteyou - I think that makes sense. I think it does all come down to style - looks clean to me. I updated this PR and removed it from the translation keys. Let me know what you think! Thanks for the clarification!

mikebcbc:
@meteyou - anything else you need from me on this PR, or is it good to go?

meteyou (approved):

meteyou:
LGTM! Thanks!

wikolii:
Hi, is there any way one could choose to have the machine name in front of the tab description?
In my case I normally just print the same part in many printers and having to wait for the popup with the full tab description to know the machine name is time consuming.

I would even choose not to have the filename on the tab description, maybe there could be some setting for that.
Thanks!

Issue #1315, opened 2023-03-20
Different between estimate time from slicer and mainsail


JackyHe398:
### Requested feature:

![image](https://user-images.githubusercontent.com/62739796/226386819-2732db82-2892-4741-b061-5fdebe7a6b1f.png)
From the photo, when (the actual time is >  the estimate time from the slicer), it shows "--" only. 

I think that maybe better if it can show how much time it has overrun. Users can have a clear concept about the time of the future print. Slicer is not always accurate. 

### Solves the following problem:

This is nothing related to bugs

meteyou:
then i would prefer to check the history, because you don't know how long this print will need in the current state

JackyHe398:
Interesting. 

I would like to ask is there any reason not to show that overrun time? Maybe from other perspective, it may not always be a good idea?

meteyou:
just my 2 cents. i personally don't see any advantage in displaying them. but if it's wanted by many, i see no reason not to add this feature.

mikebcbc:
@meteyou this issue can be closed - it's merged in

meteyou:
Thx for the ping

PR #1372, opened 2023-05-03
feat: Allow negative time estimate in slicer

mikebcbc:
## Description

This PR allows and displays a negative time in the slicer estimate when it's overrun - useful for dialing in slicer estimation and getting a clear picture of how much it went over.

#1315 

image

No post-deployment tasks.

### DIFF
--- a/src/store/printer/getters.ts
+++ b/src/store/printer/getters.ts
@@ -1006,7 +1006,8 @@ export const getters: GetterTree<PrinterState, RootState> = {
             'print_duration' in state.print_stats &&
             'current_file' in state &&
             'estimated_time' in state.current_file &&
-            state.print_stats.print_duration > 0
+            state.print_stats.print_duration > 0 &&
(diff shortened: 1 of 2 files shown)


src/store/printer/getters.ts
```diff
@@ -1006,9 +1006,7 @@ export const getters: GetterTree<PrinterState, RootState> = {
             'print_duration' in state.print_stats &&
             'current_file' in state &&
             'estimated_time' in state.current_file &&
-            state.print_stats.print_duration > 0 &&
-            state.current_file.estimated_time > 0 &&
```
meteyou, line 1010: i think the estimated_time should be checked. because if this is missing in the gcode metadata, mainsail would display a negative print_duration here.

mikebcbc: Yeah, good point - I added the check for estimated_time back in.

meteyou (approved):

PR #1375, opened 2023-05-04
fix: fix zip file timestamp

pauhull:
## Description

This PR fixes the timestamp format used for zip downloads. Previously, the date components were not padded with zeroes and the month was off by 1 because ``Date.getMonth()`` reports months from 0 to 11. This PR aligns the used timestamp format with [the one used by Moonraker](https://github.com/Arksine/moonraker/blob/6c7dfe592131eb95fd760ef2db83387966b250f0/moonraker/components/file_manager/file_manager.py#LL601C15-L601C15), making it more readable and also fixing the sorting.

Example for a previous timestamp: ``config-202344-17137.zip``
After change: ``config-20230504-170137.zip``

None


meteyou (approved):

meteyou:
thx for this PR!

PR #1376, opened 2023-05-04
fix: Make the correct notification appear on gcode file move

mikebcbc:
## Description

The wrong notification was being shown when moving a file, this PR makes sure that the "file move" toast shows when we move a file instead of the rename toast.

**Before**

https://user-images.githubusercontent.com/15934114/236298842-801fef15-c3b2-4f71-a6cd-7aa4e3017efa.mov

**After**

https://user-images.githubusercontent.com/15934114/236298867-29985bd1-417f-457e-9143-ab8f11ba620f.mov

No post deployment tasks.

### DIFF
--- a/src/store/files/actions.ts
+++ b/src/store/files/actions.ts
@@ -222,7 +222,7 @@ export const actions: ActionTree<FileState, RootState> = {
             const filename = payload.requestParams.dest
                 .substr(payload.requestParams.dest.lastIndexOf('/'))
                 .replace('/', '')
-            const sourceDir = payload.requestParams.dest.substr(0, payload.requestParams.source.lastIndexOf('/'))
+            const sourceDir = payload.requestParams.source.substr(0, payload.requestParams.source.lastIndexOf('/'))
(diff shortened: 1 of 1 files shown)


src/store/files/actions.ts
```diff
@@ -222,7 +222,7 @@ export const actions: ActionTree<FileState, RootState> = {
             const filename = payload.requestParams.dest
                 .substr(payload.requestParams.dest.lastIndexOf('/'))
                 .replace('/', '')
-            const sourceDir = payload.requestParams.dest.substr(0, payload.requestParams.dest.lastIndexOf('/'))
+            const sourceDir = payload.requestParams.dest.substr(0, payload.requestParams.source.lastIndexOf('/'))
```
meteyou, line 225:
```suggestion
            const sourceDir = payload.requestParams.source.substr(0, payload.requestParams.source.lastIndexOf('/'))
```
i think you should change both here.

mikebcbc: Resolved!

meteyou (approved):

PR #1394, opened 2023-05-18
Update Russian Localization

RuChef:
Update Russian Localization


meteyou:
Thank you for your PR!

Issue #1355, opened 2023-04-25
[FR] Codemirror Language Package for printer.cfg


forgodtosave:
### Requested feature:

A Codemirror Language Package for printer.cfg. It would enable a lot of useful features, like 
- error highlighting without "Restart&Save"  
- better code highlighting  
- auto indentation ... 

It would also create a syntax tree, which would be useful to find blocks of config to resolve #1320

### Solves the following problem:

- No static error checking in the config file 
- no auto indentation 
- flat syntax tree 

### Additional information:

I would like to try it and would start based on this example, but I’m not sure if there will be any license problems or what a good directory structure would look like.

meteyou:
pls check the open PRs. most of these things are already did...

forgodtosave:
Sorry but I can't find anything regarding this in the open PRs. There is only text highlighting for gcode but nothing regarding the printer config and a grammar for it to create a language package. Could you maybe send a link?

meteyou:
@forgodtosave ohhh that sounds really nice!

can you use this PR/branch to check the syntax highlighting for your language package -> https://github.com/mainsail-crew/mainsail/pull/1200

forgodtosave:
Sorry, but I don't think so, at least in the current state. Lezer supports [grammar nesting](https://lezer.codemirror.net/docs/guide/#grammar-nesting) and it should even be possible between a lezer parser and a stream parser. The only problem is that in #1200 the streamParserJinja2 has been removed and integrated into streamParserKlipperConfig. Of course it would be possible to nest the new parser of #1200 into the new language, but it would be quite ugly because it would be a klipper/jinja2 paser in a klipper parser. 

Another question: why did you implement a custom jinja2 streamParser? There is a [public version](https://github.com/codemirror/legacy-modes#user-content-jinja2) available.

meteyou:
This was meant as a hint that you should follow this syntax highlighting instead of the current one from Mainsail.

I have not implemented this. This was a contribution...

zellneralex:
@forgodtosave we tested the public parser but Klipper has e.g. modified start and end variable strings. Testing with my config (https://github.com/zellneralex/klipper_config) showed in many areas that the default parser failed.

BTW our custom parser is also not 100% perfect but the best we have so far.

forgodtosave:
@zellneralex ok then of course it makes sense to use the custom parser.

[...]

PR #1400, opened 2023-05-27
feat: klipper config-file language

forgodtosave:
## Description

This PR adds a codemirror language packet. This includes a lezer parser for the printer.cfg file where everything related to gcode/jinja2 is forwarded to the old parser. Furthermore, code folding for the config blocks as well as syntax error highlighting is implemented. Also a basic example of auto completion was implemented.

<del>Here the following question arose:
How context sensitive should autocomletion and linting be? 
Should only options be suggested that are also available under the following config block?
Should all options that cannot exist in the context also be marked as errors?
Should only certain data types be available after certain options?</del>

<del>This all should be possible (see auto completion example) but you would have to store all this information somehow.
Hard coding would be possible, but you would have to keep it up to date.
Of course, it would be much nicer and more effective to automatically extract this information from [this file](https://github.com/Klipper3d/klipper/blob/master/docs/Config_Reference.md) when updating klipper. That should be possible for the most part, even if certain dependencies may not be easy to extract.
I could try that, the only question is whether and how I should proceed. In an extra PR, ...?</del>
(part of other new PR)

I'm still unsure about the following things and would appreciate feedback and ideas:
- file structure (I used codemirror and lezer as inspiration)
- Code highlighting (I currently like that each data type has its own color)
- I added a script in package.json to generate the parser which needs to be done whenever the grammar is edited. Should that stay, rename, ...?
- The outsourcing works, but the external parser seems to have a problem getting only this part of the configuration file, at least the color coding is not correct.
- <del>the auto generated files throw an styling error, should I really let the prettier run over then?</del>

Solution for issue #1355


meteyou:
hey! thx for your PR. i'm looking forward to seeing when this one is finished.

- do you think it make sense to "convert/integrate" the gcode/jinja parser to lezer?
- i am currently not sure if it makes sense to mark all "unrecognized entries for this block" as errors. because new klipper settings would then also be displayed as "incorrect/error" until they are integrated into mainsail. or even if klipper is older than the mainsail version, this would not match. it would be best to get all possible variables from klipper during initializing.

meteyou:
please add these also to the `.prettierignore`:
```
src/plugins/languages/dist
src/plugins/languages/KlipperConfigLanguage/*.d.ts
```
this should fix the PR check and i think its not necessary necessary to "clean" these files.

forgodtosave:
i have now added these two lines to `.prettierignore`:
```
src/plugins/CodemirrorLanguages/KlipperConfigLanguage/parser/klipperConfigParser.terms.js
src/plugins/CodemirrorLanguages/KlipperConfigLanguage/parser/klipperConfigParser.js
```
The paths in your suggestion don't seem to fit, or should that be an idea for a new folder structure?

I have now removed the auto completion and will add it better with better error highlighting in an extra PR.
The current error highlighting should fit because as long as the naming scheme of the future klipper config blocks/options does not change dramatically, there will be no syntax errors.

meteyou:
sry for the long delay...

- Yes my paths could be wrong. I don't tested it. I just extract it from the prettier error message.
- I think its a good idea to have a separate PR for the auto completion.
- Klipper don't support all commands functions of Jinja. Only a part of it. The best way for testing, is to check AlexZ printer config: https://github.com/zellneralex/klipper_config. I think you will find all possible options in this config.
- Would it be possible to shorten the paths a bit for language files? `CodemirrorLanguages` and `KlipperConfigLanguage` are really long names.

forgodtosave:
No problem
- I shorten the paths to `src/plugins/Codemirror/KlipperCfgLang/parser/klipperCfg.grammar` is this ok?
- I found also a very strange bug: right know it is not possible to start a file with a comment (the problem is probably caused by lezer and how I import the grammar/tokenizer)  so I marked this PR as draft again

meteyou:
maybe you can also drop the "Codemirror" directory.

forgodtosave:
mh I don't know. I the new version there will be some files in Codemirror folder which are not specific to any language like printLezerTree for better debugging or the configuration for the parser testing framework. Tomorrow I will hopefully have the new version running and will push it then you can form your own opinion.

forgodtosave:
Fixed some grammar bugs.
I've also added:
- a few more value types that should make linting easier 
- some tests that should test most of the grammar rules
- test whether a parsing error occurs with any of the configurations of AlexZ
- code to `codemirror.vue` (commented out) that can be used to debug the grammar.

But I'm not sure if it wouldn't be better to only build the parser/language if changes were actually made to it, instead of doing it with every build.

meteyou:
@forgodtosave any updates here?

meteyou:
I am closing this PR because no response. Please create a new (up-to-date) PR if you want to continue working on this feature.

Issue #1398, opened 2023-05-26
Let macros explicitly define display options


jschuh:
### Requested feature:

Not sure if this belongs under #484 or not, but it would solve a number of frustrations I've been dealing with when trying to integrate [my own macros](https://github.com/jschuh/klipper-macros) with front-ends. 

My suggestion is to add a scheme where a macro exposes it's front-end presentation as an explicitly named variable containing a dictionary of fields. For most macros this will just be a static dictionary that sets default values and input ranges. For more complex macros this dictionary can be updated in response to changes in printer state.

Here's a rough example of what I'm thinking:

```
[gcode_macro MY_MACRO]
variable_front_end_options: {
  'hidden': False, # Set to true to hide
  'params': {
      'MY_SELECT_ARG' { # Drobdown box
        'type': 'select'
        'select': ['choice_a', 'choice_b', 'choice_c']
        'default': 'choice_a'
      }
      'MY_NUMBER_ARG' { # Numeric entry field
        'type': 'number'
        'min': 0
        'max': 100
      }
      'MY_TEXT_ARG' { # String entry field
        'type': 'text'
        'max': 10
        'default': 'purple'
      }
    }
  }
# More macro stuff after...
```

The general idea is that before rendering the macro parameters pane, Mainsail would check for store.state.printer['front_end_options']; if it's not present it just falls back to what it inferred from parsing the config. If it is present, then it renders only the parameters in the dictionary, and according to the specified format.

In the case of MY_MACRO above the rendering would be as follows:

- MY_SELECT_ARG as a dropdown box with the default set to "choice_a"
- MY_NUMBER_ARG as a numeric value between 0 and 100 with no default
- MY_TEXT_ARG as a text entry field accepting up to 10 characters with a default of "purple"

### Solves the following problem:

The big things that this allows for are:

- parameters that determine their default values from printer state
- parameters that can have basic restrictions
- parameters that can be hidden from the UI
- hiding a macro without the underscore prefix (e.g. for macros that make sense in the console, but not as buttons)

meteyou:
i already thought of something similar, but how you managed it with 1 variable didn't occur to me yet. i find your suggestion very well realized. Thank you!

jschuh:
Thanks. If you look at my repo you'll find I've stretched the capability of macros far beyond what any reasonable person should (and I am not exactly proud of this 😉).

[...]

PR #1401, opened 2023-05-27
feat: Let macros define display options (WIP)

jschuh:
This is a first pass at what I proposed in #1398. It's also my first time doing anything at all with vue.js, so feedback on any of that would be helpful. Here's a picture of what it looks like, and below is the macro I've been using to test:

image

So, it successfully overrides the parsed parameters with the ones from the macro variable. The big things I know I have left are:

- Make it so the macro button is not added at all if marked hidden (rather than just disabled).
- Figure out how to force re-rendering when `front_end_options` updates inside of Klipper.
- Add some documentation for the schema.
- Maybe add a description field that would populate a tooltip for each parameter.

### DIFF
--- a/src/components/inputs/MacroButton.vue
+++ b/src/components/inputs/MacroButton.vue
@@ -5,7 +5,7 @@
             :color="color"
             :class="paramArray.length ? 'macroWithParameters' : ''"
             :loading="loadings.includes('macro_' + macro.name)"
-            :disabled="disabled"
+            :disabled="disabled || paramsRequired"
             @click="doSendMacro(macro.name)">
             {{ alias ? alias : macro.name.replace(/_/g, ' ') }}
         </v-btn>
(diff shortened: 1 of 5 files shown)


meteyou:
i could think of boolean and "slider/range" as possible types for the variable. would you be interested in implementing these two types as well?

PS: I love this PR!

jschuh:
I was already planning on adding a `step` parameter as a hint to the front end that it can render as a slider, so something like `{'min':0,'max':100,'step':1}` would give a slider from 0 - 100.

I'd been trying to decide what's best for a `bool`, and I'm thinking of something like `{'bool':[0,1]}` where the iterable is just a pair with the false value first and the true value second. That should work for pretty much all the checkbox scenarios.

Relatedly, I'm thinking I should use `enum` instead of `select` for the field name in the select box case, because everything else is named for a data type rather than a control, so I should probably keep it consistent. I could even add an `enum_multiple` boolean flag to allow for multi-select.

I also spent a little bit of time reading up yesterday on vue's reactive model, since that's the only place I'm hitting significant issues. I have zero experience with vue (or really any of these reactive web toolkits) and am still unclear on how to trigger re-rendering of the UI elements when the underlying data changes. I'd appreciate any pointers on that front.

meteyou:
`step` could also be possible for number inputs. i think we should orientate the names on "gui stuff", because the user will configure the gui with it. so i think an option "slider" to have a "slider" is more intuitiv and select instead of the type enum.

meteyou:
i just checked the getMacros getter. it only loads the data from the config/settings object of klipper. this is only created once when klipper is started and is not changed afterwards. to be able to dynamically read the current values, you have to read printer['gcode_macro xyz'] and not the config. it may be that i have a short time on monday, then i could extend the getter getMacros to also have access to the current variables. this was simply not used/needed in mainsail until now.

jschuh:
That doesn't match what I'm seeing, and this line certainly gave me the impression that the macro variables are coming from the printer state: https://github.com/mainsail-crew/mainsail/blob/84814080291e6f037230491c3e273ddd49ed6e53/src/store/printer/getters.ts#L179-L181

What I'm seeing is that the macro state gets read once at page load. So, if I reload the page any changes I made to the Klipper macro variable gets propagated to the UI (without reloading Klipper). However, the UI doesn't update unless I reload the whole page.

I did also try pulling `front_end_hints` directly from `this.$store.state.printer`, but I still seemed to get stale data until I reloaded the page.

meteyou:
Ohh! You are right! I overlooked it. I have to check, if mainsail also subscribe it and if the getter works correct.

jschuh:
Yeah, I was guessing it's something along those lines. I spent quite a few hours digging into the reactive UI stuff, before I figured out that the data updates simply weren't propagating. Then I dug into that for a bit, reading up on vuex and trying to figure out how the state propagates from Moonraker. But that's all quite a bit deeper in the weeds than I was expecting to get, and I don't feel like I have a decent grasp of it.

If you have some pointers (ideally an existing pattern in the code that I could follow) then I'd be willing to take another crack at it.

meteyou:
Ok. I have added a debug output in the incoming Websocket data to doublecheck it. Moonraker doesn't send a update with your macro, but I get updates with this macro:
```yaml
[gcode_macro test_macro2]
variable_test: 0
gcode:
    {% set test = params.TEST|default(0)|float %}
    SET_GCODE_VARIABLE MACRO=test_macro2 VARIABLE=test VALUE={test}
```
I can easily remember that Moonraker only updates the printer data 2 levels deep or updates pushed. The same issue was in the bed_mesh data.

jschuh:
Gotcha. Guess I'll have poke around at the Moonraker code and get a better sense of how this is handled. Hopefully I can get it to update the state without having to do anything to ugly with macro variables.

Zeanon:
I have tried your code and encoutered two problems:
Macros with parameters are not visible in mobile mode(neither with the new param system nor with the old)
and if you have a "select" param and press the "set default arrow" the field will be blank

meteyou:
@jschuh any update here?

jschuh:
Sorry, I ended up going down the rabbitthole of figuring out how to get Moonraker to propagate the variable updates. But I hadn't quite gotten that sorted out before real life got in the way, and now I just haven't had time to circle back.

meteyou:
@jschuh thx for your answer!

if you need help somewhere, feel free to ping me. maybe we should/could also ask Arksine...

jschuh:
@meteyou, Sorry, it's really just that I haven't had time to get back to it (and probably won't until at least next month).

That stated, it would definitely be helpful to get some feedback from Arksine on propagating the variable updates from Moonraker, because that was the only major sticking point.

Arksine:
This isn't something that can be resolved in Moonraker.  This occurs due to the way Klipper creates "diffs" for status updates.  Klipper stores a reference to the dict returned by get_status(), calling `__setitem__` to update a nested  dict doesn't change the top level reference, so Klipper does not detect the change and does not push it to Moonraker.

The only way to make Klipper detect the change is to call `SET_GCODE_VARIABLE`.  This doesn't work well with dicts though, and large ones will be problematic.  It might be necessary to extend gcode_macro.py with functionality specific for this purpose.

jschuh:
Okay, I have some time to look at this again.

Creating a variable for each hint feels hacky, but it is certainly an easy enough solution. That stated, I feel like it would be useful for Klipper to have a more formal way of providing this state to frontends (similar to the EXCLUDE_OBJECTS polygons and the SET_PRINT_STATS_INFO state).

I'll take a look at how to best approach this, and post something in the next day or so.

jschuh:
I'm still thinking through what approach makes sense here and catching up on the [related discussion in Klipper PR 6393](https://github.com/Klipper3d/klipper/pull/6393).

meteyou:
@jschuh any news here?

TypQxQ:
This looks very promissing!
Any updates @jschuh

jschuh:
> @jschuh any news here?

Sorry @meteyou, I thought I'd have time to get back to this but turns out I haven't, and I don't expect that to change in the next several months.

meteyou:
@jschuh thx for the update! pls let me know, when you have time to continue the work on it.

PR #1416, opened 2023-06-14
feat: add total/selectable print time and filament usage to gcode files panel

tstone:
## Description

This PR adds status labels for total print time and total filament usage to the G-Code Files panel. 

- The initial state shows the totals for the entire directory
- Once files are selected, the totals show for only those selected

The purpose of this feature is to make multi-prints projects easier, either by using the totals at the start of a project, or selecting specific files during a project to make sure enough filament remains or to assess how much total time is remaining.

There were several places this data could go, and after 3 iterations and feedback from fellow Mainsail users, I chose to locate it in the top right. The reason for this was twofold:

- There is already an "info bar" atop the files table with a Label: Value format, making this a logical choice
- The buttons in the top right already change based on files selected, so having the totals changed on selection should be more natural as all the things change in the same location

Adding more data to the info bar created a bit of a problem for smaller, mobile screens. In this case I chose to drop the total filament usage data, keeping the total print time, in order to not overly clutter the UI.

Default state, show the total for the entire directory:
image

Selected state, showing the total for only the selected files:
image

Landscape mobile, dropping the filament usage to keep the UI clean:
image

Portrait mobile, moving values to be under their label:
image

Signed-off-by: Titus Stone <wastingtape+mainsail@gmail.com>

### DIFF
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -159,7 +159,7 @@
                         <template v-if="getColumnSum('filament_weight_total')">
                             <v-tooltip top>
                                 <template #activator="{ on, attrs }">
-                                    <span v-bind="attrs" v-on="on" class="filament-weight-sum mr-6">
+                                    <span class="filament-weight-sum mr-6" v-bind="attrs" v-on="on">
                                         <b>{{ $t('Files.FilamentWeight') }}:</b>
                                         {{ getColumnSum('filament_weight_total') }}
                                     </span>
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -719,14 +719,6 @@ export default class GcodefilesPanel extends Mixins(BaseMixin, ControlMixin) {
     mdiCheckboxMarked = mdiCheckboxMarked
     mdiDragVertical = mdiDragVertical
 
-    readonly summableColumns = Object.freeze([
-        'filament_total',
-        'filament_weight_total',
-        'estimated_time',
-        'last_total_duration',
-        'last_filament_used',
-    ]);
-
     formatFilesize = formatFilesize
     formatPrintTime = formatPrintTime
     sortFiles = sortFiles
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -1111,17 +1111,17 @@ export default class GcodefilesPanel extends Mixins(BaseMixin, ControlMixin) {
     }
 
     getColumnSum(columnValue: string) {
-        const col = this.tableColumns.find(c => c.value == columnValue);
+        const col = this.tableColumns.find((c) => c.value == columnValue)
         if (col && col.outputType) {
-            let files = this.files;
+            let files = this.files
             if (this.selectedFiles && this.selectedFiles.length > 0) {
-                files = this.selectedFiles;
+                files = this.selectedFiles
             }
 
             const sum = files.reduce((sum: number, file: FileStateGcodefile) => {
-                return sum + (file[col.value] || 0);
-            }, 0);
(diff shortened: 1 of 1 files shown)


meteyou:
Thank you for the PR! I only think about it to split the path and the "stats" in separate lines, because the path could be very long. Can you please split it in separate lines?

EmJay276:
> Landscape mobile, dropping the filament usage to keep the UI clean:

I would suggest keepping then filament usage also on mobile. Very helpful to check if enough filament is left if I want to start a job one more time.

> Default state, show the total for the entire directory

I would suggest to hide the state if no job is selected. The UI would be more clean and it can be easily shown if all are selected.
Perhaps only show 1 decimal on filament and no seconds on print time to save space.


> I only think about it to split the path and the "stats" in separate lines, because the path could be very long.

Seperate lines would take a lot of space on mobile, I would rather suggest to shorten the description
`Print Time: 5h 5min  Filament: 135.6g  Free Disk: 8.4 GB`

meteyou:
but if you only display it, when files are selected, i see no problem with a separate line. then you have also enough space to display:

```
3 files selected, Print Time: 5h 5min,  Filament: 135.6g
```
and  path & free disk space will be in the first line without any change.

EmJay276:
> but if you only display it, when files are selected, i see no problem with a separate line. then you have also enough space to display:
> 
> ```
> 3 files selected, Print Time: 5h 5min,  Filament: 135.6g
> ```
> and  path & free disk space will be in the first line without any change.

Seems fine for me 👍

tstone:
To make it a second line, there are some downsides as well:

- If the 2nd line is not initially visible, then selecting an item would move the entire UI below it down by a line height, which I could see being annoying to users, making selecting other items a bit of a "moving target".
- If the 2nd line is initially there but empty and only populated once a selection is made, it just adds a bunch of extra whitespace that I suspect would look odd (I saw this because my first implementation was actually to add an additional spreadsheet-style "Sum" row to the table, but there was way too much whitespace and it looked odd).

I had considered putting the data in the bottom left, on the same row as the pagination controls, but that would likely create some of the same issue.

A third idea I had 🤔, but haven't experimented with, was to move everything "stats" related to a new panel below the file selection panel. This panel would work similar to the way the "History" page does, where there are some graphs, and changing the selection changes the content of the graph. This would be more work, but would get out of the problem of putting too much extra information in the files table, and also allow users to collapse the panel if they didn't want that data..

I'm attaching a (really bad) quick sketch of what I'm takling about. This new panel could have a pie chart of printed vs. not printed, and a table of filament length, filament weight, and print time of the selected files. I'm sure other things could be added in the future as well.

image

meteyou:
Is adding the panel with the charts necessary?

My main problem with this is that if you don't select any files, mainsail doesn't have all the metadata, but only the metadata from the currently visible files (or metadata pulled since the last refresh). The statistics wouldn't make sense without having a file selected.

As an alternative, it would occur to me that you make the two lines above by default and then split them:

- 1. line: path
- 2. line (per default): 0 selected files ... free disk space
- 2nd line (with files selected): 3 selected files ... print time ... filament ... free disk space

EmJay276:
> As an alternative, it would occur to me that you make the two lines above by default and then split them:
> 
>     * 1. line: path
> 
>     * 2. line (per default): 0 selected files ... free disk space
> 
>     * 2nd line (with files selected): 3 selected files ... print time ... filament ... free disk space

I like this idea.

meteyou:
@EmJay276 @tstone any progess here?

EmJay276:
@meteyou 
I think@tstone is developing this PR.
But I can have a look at necessary changes.

meteyou:
@EmJay276 this would be nice. otherwise I have to implement this feature, but feel free to create a new PR with this feature.

meteyou:
Any news here?

meteyou:
i will close this PR, because of no response...

PR #1419, opened 2023-06-16
feat: Adding logic to the index to handle when the PWA can load the html from disk but not the JS and the JS call fails.

QuinnDamerell:
## Description

Mainsail uses a PWA so its data can be cached locally and it's super snappy. The PWA caches some resources on disk and is able to load them without network access. However, it doesn't cache the js files. (maybe that's a bug?)

Not caching the JS files means that the website can load from a local disk, make the JS call, and fail. When that happens, the user is left with a blank black screen.

This change adds logic to detect if the main JS file has failed to load and shows the user an error message, so they know what's going on.

To repo the black screen due to JS failing, do the following:

1. Load Mainsail as normal from any local server.
2. Give the page a few seconds to allow the PWA to cache resources locally.
3. Use the browser's debug tools to disable caching to web requests (in the network tab, at the top)
4. Stop the Mainsail server so the browser can't connect to it.
5. Refresh the page.

You will see that the HTML and CSS load from the service worker, but the JS fails to load. If you don't disable the caching of the browser the JS will continue to load due to the browser caching it. But if you were to close and re-open the browser, the browser will no longer have the cache. 

None

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,36 +24,34 @@
             // need to be pulled from the server. If the server is offline or we can't connect, the JS file load fails,
             // the website just shows a black, blank, screen. The small script below binds to all of the script tags and if
             // any fail to load, add some text to the main body to indicate something is wrong.
-            var hasShownLoadFailedError = false;
-            var hasMainScriptLoaded = false;
-            function OnScriptLoadFailed()
-            {
-                if(hasShownLoadFailedError) {
-                    return;
+            var hasShownLoadFailedError = false
+            var hasMainScriptLoaded = false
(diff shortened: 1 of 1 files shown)


meteyou:
hey! thx for your PR and the hint. i double-checked the PWA caching and it is incomplete. i will finish it and make a PR for it to fix this issue.

QuinnDamerell:
Sounds great @meteyou. When you make the PR to fix it, if you think about it, tag me. I'm just interested to see the fix.

PR #1420, opened 2023-06-17
feat: add bed aspect ratio to heightmap graph

EmJay276:
## Description

This PR adds  the bed aspect ratio to the heightmap graph.
This is espacially usefull for highly non square print beads (mine is 600 x 300), see Screenshots below.

None

### Mobile
<td valign="center"> </td>
<td valign="center"> </td>

No

Signed-off-by: Michael Jäger michael@mjaeger.eu

### DIFF
--- a/src/pages/Heightmap.vue
+++ b/src/pages/Heightmap.vue
@@ -629,6 +629,8 @@ export default class PageHeightmap extends Mixins(BaseMixin, ControlMixin) {
                         },
                     },
                 },
+                boxWidth: 100 * this.scaleX,
+                boxDepth: 100 * this.scaleY,
             },
             series: this.series,
         }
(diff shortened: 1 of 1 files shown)


meteyou:
Thx for your PR!

EmJay276:
Thanks for reviewing it that fast 👍

If min(XYRange) is 0 (this can only be the case if one direction has only 1 be neat point), isn't there a problem with a 0 wide/depth graph? Even if we are not dividing by 0?

0/1 = 0

I will check later and open a PR if needed.

meteyou:
Thx for your hint! I had not thought about this case at all. It is very unusual, but we should merge a fix if it could happen.

EmJay276:
I tried to create a bed mesh with a X/Y distance of 0, but klipper won't allow this kind of settings (see below).
I think we don't need to check this case, if the bed with id 0, we have other problems than a heightmap graph not showing up.

Only 1 point along an axis is invalid: `Option 'probe_count' in section bed_mesh must have a minimum of 3`
Setting min/max points to the same value is also invalid: `bed_mesh: invalid min/max points`

I also noticed, having the mesh on a part of the bed, the graph will still ajust to the size of the bed (which was it doing the same way before but without axis scaling).
Which is perfectly fine I think.

Mesh with 290 x 290, bed 300 x 600

PR #1437, opened 2023-06-23
feat: add nevermore sensor support

SanaaHamel:
Support for sensors added by [Nevermore controllers](https://github.com/SanaaHamel/nevermore-controller).

Signed-off-by: Sanaa Hamel <sanaahamel@outlook.com>


SanaaHamel:
I'd prefer a non-hardcoded solution, but this one is the smallest change that solves the issue.

meteyou:
i think we can close this PR, because of the new implementation. Thx @SanaaHamel for this initial PR.

PR #1524, opened 2023-08-19
feat(move): implement #1209

### LINKED ISSUE #1209: Gcode file page - Move to & Move to with multiple selected
### Requested feature:

 On the gcode file page there is currently no way to efficiently move files into different folders.  

The only option is drag and drop, but what if the file is on a different page? Also, if the file is farther down in the list there is no obvious way to scroll the page up while dragging the file.

It would be nice to have a way to efficiently move files into folders including the ability to move multiple selected files at once.

 On the gcode file page there is currently no way to efficiently move files into different folders or folders that are on a different part of the page, or different page? 

### Additional information:

Thank you for everything you do!!!   ps....did you like that reminders based on print time idea? ;)

intothevoid1986:
## Description

This PR adds the feature reqested in FR #1209.
Code is recycled from "rename" function, properly relabelled to improve UX.

The feature impact ConfigFilePanel, GCodeFilePanel and Status/GcodefilesPanel

Fixes #1209 

### Desktop view GCode Files

Schermata del 2023-08-19 14-19-51

### Desktop view Config Files

Schermata del 2023-08-19 14-21-32

### Mobile view GCode Files

Schermata del 2023-08-19 14-20-34

### Mobile view Config Files

Schermata del 2023-08-19 14-21-02

### Move-to Dialog

Schermata del 2023-08-19 14-22-02


meteyou:
Sorry for the late response.

I think you misunderstood the FR. In this PR, you added the "rename" function with new names again. But the user wanted to select several files and move these simultaneously to different directory. In the background, the Moonraker "move" API is used, but it should not change the name but the location.

meteyou:
i will close this PR, because of no response...

PR #1532, opened 2023-08-26
feat: add monitors (like TMC2240) to Temperature Panel

meteyou:

## Description

This PR add support for "monitors" (like TMC2240) and an option to hide all these values in the Temperature Panel

none

image


src/components/panels/Temperature/TemperaturePanelList.vue
```diff
@@ -93,6 +112,9 @@ export default class TemperaturePanelList extends Mixins(BaseMixin) {
                 // hide MCU & Host sensors, if the function is enabled
                 if (this.hideMcuHostSensors && this.checkMcuHostSensor(fullName)) return false
 
+                // hide monitors, if the function is enabled
+                if (this.hideMonitors && this.available_monitors.includes(fullName)) return false
```
rackrick, line 116: remove unneeded filter condition.
We're filtering the available_sensors here. This will never be true because there is no available_monitors entry in this array.

meteyou: i removed the filter...

rackrick (approved):

rackrick (approved):

PR #1566, opened 2023-09-21
fix: add port to webcam url if port is not 80

dictor93:
This PR fixes the URL for MJPEG adaptive streamer. When used port different then th "80". (Relevant current port added to the URL)


meteyou:
i don't think, that this will work for all setups. i think this PR will also change the port for absolute paths, because it doesn't check if the URL is absolute or relative.

dictor93:
> i don't think, that this will work for all setups. i think this PR will also change the port for absolute paths, because it doesn't check if the URL is absolute or relative.

As you can see URL constructor used with second parameter (baseUrl) https://github.com/mainsail-crew/mainsail/pull/1566/files#diff-fdc7a161a3e6714807dbdb170673c137562833dde75941a4b98626f4a6187d55R97

So the url can not be relative. Is not it?

dictor93:
I really can't understand what the problem is that the request for camera data goes to the same port as all other requests in the application. That is, if my reverse proxy server for example listens on port 8081 and refers to mainsail, then all requests go to <hostname>/8081/route... so why should the cam requests go to <hostname>/webcam/ and not at <hostname>/8081/webcam ? Is there a security issue here? No, because I use a reverse proxy with authorization. Why are you @meteyou  contradicting?

Also, some of the streamer components reference the current port:  [for example
](https://github.com/mainsail-crew/mainsail/blob/develop/src/components/webcams/WebrtcCameraStreamer.vue#L42)
So, also i guess the same changes need to be implemented for Uv4lMjpeg and Mjpegstreamer

meteyou:
Yes, this is your special case, but when a user don't have the webcam on the same host or esp32-cams, you cannot overwrite your Webinterface port on the webcam address.

meteyou:
I just now checked the link to the WebRTC CameraStreamer. Yes, this streamer is wrong.

If you already do this fix, please fix all streamers for your use case.

dictor93:
> Yes, this is your special case, but when a user don't have the webcam on the same host or esp32-cams, you cannot overwrite your Webinterface port on the webcam address.

That is, you mean that my edit will work fine if the camera is on the same host as the mainsail, but there will be problems for cameras from other hosts?



> I just now checked the link to the WebRTC CameraStreamer. Yes, this streamer is wrong.
> 
> If you already do this fix, please fix all streamers for your use case.

Do you want me to add fixes to my PR? Which exactly? Remove port override?

meteyou:
> That is, you mean that my edit will work fine if the camera is on the same host as the mainsail, but there will be problems for cameras from other hosts?

yes, especially if another port is used, it is always overwritten.

> Do you want me to add fixes to my PR? Which exactly? Remove port override?

it would be nice, if you fix all webcams the same way and not only 1 client.

meteyou (approved):

PR #1575, opened 2023-10-01
build: update toolchain to the latest version

pataar:
## Description
This updates all dependencies to the latest version where possible. Except for some major updates which require too much work right now.

Note that most of the changes are due prettier going to version 3. And ESLint that's updated to the latest version

### DIFF
--- a/.github/workflows/auto-analyze.yml
+++ b/.github/workflows/auto-analyze.yml
@@ -12,7 +12,7 @@ jobs:
   auto-analyze:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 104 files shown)


meteyou:
thx for your PR! would it be possible, to change the rule with the , at last char of each attribute list? i think the most changes in all these files is this rule.

pataar:
Should be possible! I'll check it tomorrow

pataar:
@meteyou I reverted the trailing commas

meteyou (approved):

PR #1580, opened 2023-10-04
feat: light mode ui

### LINKED ISSUE #1297: Theme Switch
### Requested feature:

A switch to change the Theme from dark to Light and backwards.

I made a Vuetify3 Project as example.

code

Maybe you can also add a setting to set the colors for the Dark and Light Theme.
I uploaded a zip with the Vue-File.
[Switch-Test.zip](https://github.com/mainsail-crew/mainsail/files/10900195/Switch-Test.zip)

_No response_

### Additional information:

_No response_

inhabitworker:
## Description

Added a light mode with accompanying option under UI Settings, relying on vuetify defaults (removing explicit "dark" attributes in code etc.) and a theme mixin used to close the gap/centralize to extend consistently to chart options and other specific places where needed

Fixes #1297

image
image
image


src/components/TheConnectingDialog.vue
```diff
@@ -28,7 +28,7 @@
                 </div>
             </v-card-text>
             <v-card-text v-else class="pt-5">
-                <v-progress-linear color="white" indeterminate></v-progress-linear>
```
meteyou, line 31: Please only remove this only in light mode. So right now, this progress-line will displayed in primary color. at start up, it will be everytime blue (because the gui settings are loaded later). i would prefer a gray color in light mode and white in dark mode.

for always light mode, we also need a setting in the config.json.

src/components/TheSidebar.vue
```diff
@@ -90,14 +94,17 @@ export default class TheSidebar extends Mixins(NavigationMixin, BaseMixin) {
...
+            ...(this.boolNaviTemp && { top: `${topbarHeight}px !important` }),
+            ...(this.boolNaviTemp && { paddingBottom: `${topbarHeight}px` }),
         }
-        return {}
+
+        return vars
+    }
+
+    get backgroundImageStyle(): CSSProperties {
+        return !this.$vuetify.theme.dark ? { filter: 'invert(1)' } : {}
```
meteyou, line 107: this would also invert custom background images. pls just change the default/fallback background image, or check here, if this fallback image is used.

src/components/TheSidebar.vue
```diff
@@ -90,14 +94,17 @@ export default class TheSidebar extends Mixins(NavigationMixin, BaseMixin) {
...
 
-    get sidebarCssVars(): any {
-        if (this.boolNaviTemp) {
-            return {
-                top: `${topbarHeight}px !important`,
-                'padding-bottom': `${topbarHeight}px`,
-            }
+    get sidebarCssVars(): CSSProperties {
+        let vars: CSSProperties = {
+            ...(this.boolNaviTemp && { top: `${topbarHeight}px !important` }),
```
meteyou, line 99: I think this will always be "true", because the "condition" with the object.

src/components/TheTimelapseRenderingSnackbar.vue
```diff
@@ -2,12 +2,12 @@
 
 <template>
     <div>
-        <v-snackbar v-model="boolShowDialogRunning" :timeout="-1" :value="true" fixed right bottom dark>
+        <v-snackbar v-model="boolShowDialogRunning" :timeout="-1" :value="true" fixed right bottom>
```
meteyou, line 5: I think this will be wrong, if you just delete the dark option with no condition.

src/components/TheTimelapseRenderingSnackbar.vue
```diff
@@ -2,12 +2,12 @@
...
             <div>{{ $t('Timelapse.TimelapseRendering') }}...</div>
             <v-progress-linear v-if="progress > 0" class="mt-2" :value="progress" indeterminate></v-progress-linear>
             <v-progress-linear v-if="progress === 0" class="mt-2" indeterminate></v-progress-linear>
         </v-snackbar>
-        <v-snackbar v-model="boolShowDialogSuccess" :timeout="5000" :value="true" fixed right bottom dark>
+        <v-snackbar v-model="boolShowDialogSuccess" :timeout="5000" :value="true" fixed right bottom>
```
meteyou, line 10: I think this will be wrong, if you just delete the dark option with no condition.

src/components/TheEditor.vue
```diff
@@ -56,7 +56,7 @@
                 </v-card-text>
             </panel>
         </v-dialog>
-        <v-snackbar v-model="loaderBool" :timeout="-1" :value="true" fixed right bottom dark>
+        <v-snackbar v-model="loaderBool" :timeout="-1" :value="true" fixed right bottom>
```
meteyou, line 59: I think this will be wrong, if you just delete the dark option with no condition.

src/components/TheTopbar.vue
```diff
@@ -70,7 +70,7 @@
             <the-settings-menu />
             <the-top-corner-menu />
         </v-app-bar>
-        <v-snackbar v-model="uploadSnackbar.status" :timeout="-1" :value="true" fixed right bottom dark>
+        <v-snackbar v-model="uploadSnackbar.status" :timeout="-1" :value="true" fixed right bottom>
```
meteyou, line 73: I think this will be wrong, if you just delete the dark option with no condition.

src/components/TheUpdateDialog.vue
```diff
@@ -17,7 +17,7 @@
 
 <template>
     <v-dialog :value="application !== ''" persistent max-width="800" class="mx-0">
-        <v-card dark :loading="!complete">
+        <v-card :loading="!complete">
```
meteyou, line 20: I think this will be wrong, if you just delete the dark option with no condition.

src/components/panels/Machine/ConfigFilesPanel.vue
```diff
@@ -512,7 +512,7 @@
             </panel>
         </v-dialog>
 
-        <v-snackbar v-model="uploadSnackbar.status" :timeout="-1" :value="true" fixed right bottom dark>
+        <v-snackbar v-model="uploadSnackbar.status" :timeout="-1" :value="true" fixed right bottom>
```
meteyou, line 515: I think this will be wrong, if you just delete the dark option with no condition.

src/components/panels/StatusPanel.vue
```diff
@@ -88,7 +88,7 @@
                 </v-container>
                 <v-divider class="mt-0 mb-0"></v-divider>
             </template>
-            <v-tabs v-model="activeTab" fixed-tabs dark>
+            <v-tabs v-model="activeTab" fixed-tabs>
```
meteyou, line 91: I think this will be wrong, if you just delete the dark option with no condition.

src/store/gui/types.ts
```diff
@@ -97,6 +97,7 @@ export interface GuiState {
     presets?: GuiPresetsState
     remoteprinters?: GuiRemoteprintersState
     uiSettings: {
+        theme: string
```
meteyou, line 100: pls use both options instead of string as datatype

src/store/server/history/types.ts
```diff
@@ -37,7 +37,4 @@ export interface ServerHistoryStateAllPrintStatusEntry {
         borderWidth: number
         borderRadius: number
     }
-    label: {
-        color: string
```
meteyou, line 41: why is this not used anymore?

src/App.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <v-app dark :style="cssVars">
+    <v-app :style="cssVars">
```
meteyou, line 2: I think this will be wrong, if you just delete the dark option with no condition.

src/App.vue
```diff
@@ -195,6 +199,11 @@ export default class App extends Mixins(BaseMixin) {
         })
     }
 
+    @Watch('theme')
+    themeChanged(newVal: string): void {
+        this.$vuetify.theme.dark = newVal == 'dark' ? true : false
```
meteyou, line 204: would it be possible to load the default value from the config.json at startup?

meteyou:
Thanks for your PR! This is a very popular feature. I just found some small issues:

- In dark mode, the "toolbar" and the "content" of each panel have different background color and you see the there different areas. For example in your screenshot above from the "Interface Settings", the sidebar will scroll and cut by a white line, where the toolbar starts. I would prefer a border bottom of the toolbar like here in my screenshot:
(I just added it in the chrome dev tools with `border-bottom: 1px solid #ccc;`)
image

- the snackbar is right now in "dark mode":
image

- history panel creates errors because of the changes from the charts

i also add some comments to the code changes. I will be happy to merge this feature then.

inhabitworker:
I will make changes according to your feedback. Apologies as I am new to contributing, I'm wondering if I should be creating some test(s) here, rather than eyeballing? Truthfully I am unsure when it comes to this kind of visual feature.  Thank you for reviewing my code.

src/store/server/history/types.ts
inhabitworker: The charts are given an option object, which includes both data and styling options, in some places the styling was hard coded and not easy to modify according to theme. so I cut such properties out of the object & type to be able to dynamically pass in colour/styling.

inhabitworker:
Sorry for the delay, I will list some of the changes I have made: 

- Returned "dark" attributes, with conditional check. Still not sure that these are necessary (they dictate whether to use dark light stylings within theme, which is what is being adjusted anyway).
- Initial loading of config.json is made to be synchronous before vue is mounted so there isn't an ugly transition when theme is being resolved. As such, theme is resolved further in this process too (default possibly read from file, potential user setting fetched from moonraker, falling back to dark)
- Some adjustments to charts on the history page to make sure they properly refresh with theme switches.
- "snackbars" themed, "dragable" elements in settings also themed, which were previously themed by hard coded css class, now as style.
- border-bottom on panel title bar (toolbar?)

There's a few little changes here and there, I tried to tie things once again into the theme mixin, so these visual settings can be more easily located in case of other developments

src/main.ts
```diff
@@ -75,24 +76,33 @@ const initLoad = async () => {
...
             window.console.debug('Loaded config.json')
 
-            await store.dispatch('importConfigJson', file)
+            store.dispatch('importConfigJson', file)
             if ('defaultLocale' in file) i18n.locale = file.defaultLocale
+
+            // Handle theme outside of store init and before vue mount for consistency in dialog
+            let theme = file.defaultTheme ?? defaultTheme
+            const uiSettingsUrl =
+                store.getters['socket/getUrl'] + '/server/database/item?namespace=mainsail&key=uiSettings'
```
meteyou, line 85: This will not work, because in some modes, you don't have a socket URL right now. you can only store a default theme setting in the config.json itself.

for example in my.mainsail.xyz. here will be only load the setting "instanceDB: browser" here without any moonraker connection. only after add/select a printer in the printer select dialog, there exists a moonraker url.

meteyou:
sry for the late response. can you doublecheck if the dark option is needed? if it is not needed, i would prefer to remove it complete.

i also added a comment to the "default setting load". this cannot work this way, because at this moment, its possible, that there don't exist a moonraker url.

meteyou (approved):

meteyou (approved):

h3rm:
it would be cool if /assets/index-02519ded.css in mainsale/files wouldn't have this row:
`#page-container {
#   max-width:1800px`

Without manipulating the css no horizontal scrolling is possible & there is no horizontal scroll bar:

2024-02-24 05_24_57-14% Printing - ETA_ 21_16 - mt-v2-vice-work-bench_PLA_10h0m gcode- v400 – Mozill

with "max-width" commented out in the css it looks much better because suddenly there are >15 columns visible instead of <10 :) :
2024-02-24 05_24_19-14% Printing - ETA_ 21_16 - mt-v2-vice-work-bench_PLA_10h0m gcode- v400 – Mozill

can you please adapt the css? the better solution than deactivating "max-width" would be horizontal scroll bar...

PR #1602, opened 2023-10-10
feat: add settings for public url of the spoolman frontend

Roguyt:
## Description

This allow to specify the real IP or domain of the spoolman client which is used in the Mainsail UI to open it through the two buttons "Open Spool Manager"

This resolves an issue I have since my spoolman instance is set as 127.0.0.1 for moonraker but this isn't obviously the good link to access the Web UI.

image

No.


meteyou:
@Roguyt would it be also possible to just replace 127.0.0.1/localhost with the current moonraker connection host?

PR #1630, opened 2023-10-28
feat: add action command prompt dialog

### LINKED ISSUE #1625: Add support for action command prompts
### Requested feature:

Octoprint has had support for [Action Command Prompts](https://docs.octoprint.org/en/master/bundledplugins/action_command_prompt.html) for quite some time now. 

It would be great if Mainsail also supported it.

They are very useful for user interactive macros, e.g. filament loading / unloading, color change, confirmation prompts, etc.
[See this very elaborate example](https://gist.github.com/unraze/801cb0402b67635d7566b2a1a3688a24) (it says `for Fluidd` in the description, although I'm not sure if Fluidd supports this, Octoprint definitely does.)

### Additional information:

The protocol is quite simple and described on the linked octoprint page.

meteyou:

## Description

This PR is right now only a concept...

fixes #1625 

none

none

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -17,6 +17,7 @@
             <the-manual-probe-dialog />
             <the-bed-screws-dialog />
             <the-screws-tilt-adjust-dialog />
+            <the-action-command-prompt />
         </template>
         <the-select-printer-dialog v-else-if="instancesDB !== 'moonraker'" />
         <the-connecting-dialog v-else />
@@ -41,11 +42,13 @@ import TheManualProbeDialog from '@/components/dialogs/TheManualProbeDialog.vue'
 import TheBedScrewsDialog from '@/components/dialogs/TheBedScrewsDialog.vue'
 import TheScrewsTiltAdjustDialog from '@/components/dialogs/TheScrewsTiltAdjustDialog.vue'
 import { setAndLoadLocale } from './plugins/i18n'
+import TheActionCommandPrompt from '@/components/dialogs/TheActionCommandPrompt.vue'
 
 Component.registerHooks(['metaInfo'])
 
 @Component({
     components: {
+        TheActionCommandPrompt,
         TheTimelapseRenderingSnackbar,
         TheEditor,
         TheSelectPrinterDialog,
--- a/src/components/console/ConsoleTableEntry.vue
+++ b/src/components/console/ConsoleTableEntry.vue
@@ -1,10 +1,7 @@
 <template>
-    <v-row :class="'ma-0 ' + entryStyle">
+    <v-row :class="entryStyle">
         <v-col class="col-auto pr-0 text--disabled console-time">{{ entryFormatTime }}</v-col>
-        <v-col
-            :class="colorConsoleMessage(event) + ' ' + 'console-message'"
-            @click.capture="commandClick"
-            v-html="event.formatMessage"></v-col>
+        <v-col :class="messageClass" @click.capture="commandClick" v-html="event.formatMessage" />
     </v-row>
 </template>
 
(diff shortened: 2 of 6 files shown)


ulope:
Awesome!
That was really quick!

The implementation does seem to deviate a bit from the [octoprint "spec"](https://docs.octoprint.org/en/master/bundledplugins/action_command_prompt.html#supported-action-commands) though.

Octoprint "protocol":
- Commands:
  - `begin <title>`
  - `(choice|button) <text>` - those are just aliases of each other
  - `show`
  - `end`
- Response: `M876 S<index-of-choice>` (the G-code used is configurable).


While in this PR it looks like (as far as I could infer from the code):
- Commands:
  - `begin <title>`
  - `button <text>[[|command]|color]` - no `choice` here
  - `button_group_start`
  - `button_group_end`
  - `text`
  - `show`
  - `close` - vs. `end` above
- Response: `<command>` (or the text??, not sure about the code here)

I like the extensions since they allow richer interactions, but I think the basic protocol should remain compatible with the one from octoprint.

Phil1988:
Very nice!
This will level up the user experience and make the interface even more usable.
Cant wait to have this integrated  👍

meteyou:
@ulope both are not compatible. so you have to write your macros complete different, that you can use this in klipper, because klipper has no "wait for an user input" function and the prompt_end command don't definite the "end" of the declaration. so the "close" command is really a "close" of the prompt.

I already wrote the docs for this feature. maybe it will be clearer with that:
screencapture-docs-mainsail-xyz-changes-WqGAJH0GaI5torSkb3GO-overview-features-macro-prompts-2023-10-30-21_38_46

ulope:
> @ulope both are not compatible. so you have to write your macros complete different, that you can use this in klipper, because klipper has no "wait for an user input" function and 

Well you don't have to.
[Take a look at the M600 macro](https://gist.github.com/unraze/801cb0402b67635d7566b2a1a3688a24) I linked in the original issue. Even on klipper it's possible to write macros that use the a singe response dispatch gcode.

But it's probably right that the way you implemented it is much easier to use for Klipper users...

> the prompt_end command don't definite the "end" of the declaration. so the "close" command is really a "close" of the prompt.

So is `end`:

>> **prompt_end**
>> Tells OctoPrint that the dialog should now be closed (e.g. the user made the choice on the printer directly instead of through OctoPrint). Will be ignored if no dialog is defined yet.

zellneralex:
@ulope: I prepared for our testing a flow that contains M600 and filament runout. Doing that gets pretty “ugly” fast.
That was one reason why we deceived to simplify the process for most user flows.

alao the use of Octoprint and moonraker/mainsail is nothing that is done in parallel so that there is in my opinion no need to be 100% compatible to the octoprint version.

zellneralex:
choice/button: we believe the word button is the better here and why having 2 cmds that do exactly the same
end/close: 
in the octo world end really close the dialog box and ends the waiting (gcode streamer)
in the Moonraker/mainsail world this really only closes the box, if you do net set a PAUSE in the macro calling the box the print would continue.

ulope:
> @ulope: I prepared for our testing a flow that contains M600 and filament runout. Doing that gets pretty “ugly” fast. That was one reason why we deceived to simplify the process for most user flows.
> 
> alao the use of Octoprint and moonraker/mainsail is nothing that is done in parallel so that there is in my opinion no need to be 100% compatible to the octoprint version.

Yeah, I've been working on a PR that adds support for prompts to KlipperScreen and after using it a bit I agree that the model you came up with will lead to much simpler and easier to write macros. 

My only concern regarding compatibility is that people may want to switch from / to octoprint  and then would have to re-work their macros. 
But that may not be as big of a concern in practice...

zellneralex:
> My only concern regarding compatibility is that people may want to switch from / to octoprint and then would have to re-work their macros. But that may not be as big of a concern in practice...

In my opinion that will not be an issue. They need to rework there Octoprint command flow anyhow. Look at the M600 example
that [gcode_macro M876] must be always added and developed going from Octoprint to moonraker/mainsail

I hope you using the same command we do to give the user a possibility to see and do the same regardless what UI he uses in that moment.

ulope:
> I hope you using the same command we do to give the user a possibility to see and do the same regardless what UI he uses in that moment.

Yes, that was my initial concern to have alignment across all implementations, but as I said above, I've come around to your point of view in the meantime.

meteyou:
@ulope plus post your discord name, or write me a pn. I will add you to a channel.

ulope:
> @ulope plus post your discord name, or write me a pn. I will add you to a channel.

Cool, my username is `@ulope_`

Zeanon:
> @ulope both are not compatible. so you have to write your macros complete different, that you can use this in klipper, because klipper has no "wait for an user input" function and the prompt_end command don't definite the "end" of the declaration. so the "close" command is really a "close" of the prompt.
> 
> I already wrote the docs for this feature. maybe it will be clearer with that: screencapture-docs-mainsail-xyz-changes-WqGAJH0GaI5torSkb3GO-overview-features-macro-prompts-2023-10-30-21_38_46

You can actually create pseudo asynchronity in klipper by having a starting macro which sets everything up, then activates a delayed gcode that calls itself while checking a variable, then you have a macro which sets that variable and then the delayed gcode reads it and continues.
I know it's convoluted and not the easiest way but it works

meteyou:
> > @ulope plus post your discord name, or write me a pn. I will add you to a channel.
> 
> Cool, my username is `@ulope_`

sry. i cannot find this user on the Mainsail discord....

ulope:
> sry. i cannot find this user on the Mainsail discord....

I’m definitely there. Sent you a PN on discord.

rackrick (approved):

PR #1645, opened 2023-11-10
locale(fr): Add translation clean_nozzle and purge_filament

Bahoue:


meteyou:
Thx! For your PR!

The issue with the PR check is my fault. I have to fix the workflow.

PR #1656, opened 2023-11-15
feat: show filament sensor state even when the sensor is disabled

EiNSTeiN-:
## Description

This PR changes the FilamentSensor view so the filament state (detected or empty) is shown regardless of whether the sensor is enabled or disabled. I added the "disabled" state next to the presence state simply because I'm unaware of an alternative way to show both together, let me know if there is an existing pattern I can re-use here.

I sometimes disable my filament motion sensors but I would still like to see whether the sensor reports motion or not. This can be useful when debugging filament run-out issues, for example when the detection length is too short and the filament sensor keeps triggering but the print is actually going fine.

Screenshot from 2023-11-15 16-09-04

### DIFF
--- a/src/components/inputs/FilamentSensor.vue
+++ b/src/components/inputs/FilamentSensor.vue
@@ -1,9 +1,3 @@
-<style scoped>
-._filamentRunout-subheader {
-    height: auto;
-}
-</style>
-
 <template>
     <v-container class="px-0 py-2">
         <v-row>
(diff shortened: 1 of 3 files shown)


meteyou:
@EiNSTeiN- are you sure that the status will be updated even if it is disabled? i'm not sure right now and would have to reproduce and test it. i think it was a problem back then why we didn't implement it.

EiNSTeiN-:
> are you sure that the status will be updated even if it is disabled?

Yes I verified this happen, you can see in the screenshot too that the status is `detected` even though the sensor is disabled. The relevant code is linked below, you can see `self.filament_present` is always updated and `self.sensor_enabled` is checked 2 lines below and is only used to determine whether the runout gcode is executed or not.

https://github.com/Klipper3d/klipper/blob/795ce490a05b2ebae724cff634317f38bc38d721/klippy/extras/filament_switch_sensor.py#L65

What you may be remembering is the fact that the sensor status is not updated when the filament is not moving. But as long as the filament is moving, for instance whenever the printer is printing, then the internal status is updated regardless of whether the sensor is enabled or not.

meteyou:
Thx for your PR! i only fixed some issues in this module (some older issues i found while reviewing it. not your fault). i also removed the disabled text, because it is already visible with the switch itself. and now the text "empty" is also colored. this was a bug of the old code (so my fault)

PR #1670, opened 2023-12-05
feat: add minimum_cruise_ratio support in MotionSettingsPanel

meteyou:
## Description

This PR switch from `max_accel_to_decel` to `minimum_cruise_ratio`. The old field will be used as fallback, if no `minimum_cruise_ratio` exists.

none

image

Klipper PR for this feature: https://github.com/Klipper3d/klipper/pull/6418


dw-0 (approved):

b00gie-dev:
Just a little thought from UX perspective. As this value now is s ratio. Wouldn’t it make sense to expose this as a percentage? So not 0.5 but 50%?
Just a thought though. 

Greets,
Niko

meteyou:
@b00gie-dev, thanks for this idea. It sounds reasonable to me. Pls create a new FR in the issue tracker, so that I don't forget it.

PR #1678, opened 2023-12-08
feat: Improve contrast of job queue items count

mdziekon:
## Description

Job queue tab has a counter displaying current items count. The counter itself works fine, however it's presented as a gray text, which blends with the background and neighboring text. This quite poor contrast might lead to poor reception of the counter functionality (eg. for me, I usually forget that I've put something in the queue, because the counter is nearly indistinguishable from the tab's label).

To improve the "contrast" (so to speak) of the feature, a simple Badge component can be used instead of text interpolation with the tab's title, allowing it to be displayed slightly different than the neighboring text (the "contrast" improvement comes mainly from the different element being used rather than just color change). This approach also allows it to change its background depending on how many items we've placed in the queue (`gray` for `0`, `blue` for `1+`; `blue` chosen as information color).

This proposal (in a form of a ready made Pull Request) implements this change and cleans up any localisation files from previous interpolation formatting.

N/A

### Before:

before_empty_queue
before_populated_queue
before_populated_queue_open

### After:

after_empty_queue
after_empty_queue_open
after_populated_queue
after_populated_queue_open

No

### DIFF
--- a/src/components/panels/StatusPanel.vue
+++ b/src/components/panels/StatusPanel.vue
@@ -92,7 +92,7 @@
                 <v-tab v-if="current_filename" href="#status">{{ $t('Panels.StatusPanel.Status') }}</v-tab>
                 <v-tab href="#files">{{ $t('Panels.StatusPanel.Files') }}</v-tab>
                 <v-tab href="#jobqueue">
-                    <v-badge :color="jobQueueBadgeColor" :content="jobsCount.toString()">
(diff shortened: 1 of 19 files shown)


meteyou (approved):

meteyou:
@mdziekon thx for the PR. I only changed the badge to inline, because it was very close to the top border of the tab.

PR #1684, opened 2023-12-13
feat: pin files to top of gcode files list

### LINKED ISSUE #1257: Pin files to top of gcode files list.
### Requested feature:

Would be nice to have the ability to pin frequently used files to the top of the gcode files list. Then maybe an option to have this be applied on the the list page, the preview on the main page, or both.

Hunting...

### Additional information:

Sorry for all the requests! Looking forward to the next update :)

4rnoP:
## Description

Add the possibility to pin gcode file to the files view and file panel in dashboard
The pin/unpin action can be done through right click, and the icon of the file change to a pin icon to show that a file is currently pinned (while avoiding too much space).
The saved files list is store in the Moonraker DB under `gcodefiles.pinnedFiles`.

However, I'm not sure of the implementation design.
My idea is to add a property `isPinned` to the `FileStateGcodefile[]` that the panel get, and sort the files according to this property in the `sortFiles` function in `plugins/helpers`.
I placed the actions/mutations in the `store/gui/` as the `gcodefiles` db index was already accessible from there, and I also modified the getter `getGcodeFiles` to add the property to the FileStateGcodefile objects it returns, as well as the `getDeleteFile` action to remove the file from the pinnedFiles list (to avoid that if you delete a pinned file, then upload a file with the same name, it is automatically pinned). 

Fixes #1257

image
image
image


meteyou:
hey @4rnoP ! I merged this PR (https://github.com/mainsail-crew/mainsail/pull/1678). pls update your branch. what do you think about icons instead of the text labels in the tabs? i would also prefer a separate tab for pinned files.

SwiftingSpeed:
How does it handle gcode files with preview thumbnails? The ones you have in your examples all have the default file icon so maybe you haven't thought of the thumbnails since you aren't using them yourself?

Are thumbnails replaced with the pin icon as well or are they layered? Some users may have usage for the thumbnails, removing them may not be the best option.

4rnoP:
Indeed, with the current solution, the pin icon have precedence over the preview thumbnail.

However, with the separated tab for pinned files in the main dashboard, it isn't necessary to change the file icon/preview there.
For the file panel, how should the pinned files stand out if you still want the preview thumbnails visible ?

One idea is to add a small pin icon in the top right corner of the icon, like the following example
image

meteyou:
@4rnoP sry for the late response...

maybe a small icon before the filename? this would also be very prominent and easy to see.

meteyou:
I close this PR, because of no response. Pls create a new PR, when you continue to work on it.

PhilBaz:
@4rnoP Keep going bro...this would be an awesome feature. Hope you can find time :)

Issue #1618, opened 2023-10-21
Job Queue Organization


leitenky:
### Requested feature:

It would be nice if a user was able to add multiple job to the Job Queue and then be able to rearrange/resequence them.

### Solves the following problem:

The need to print items in a different order than added to the Queue

meteyou:
fixed with the PR #1692

Issue #1379, opened 2023-05-06
Allow to start any print from queue not only first one


ristein:
### Mainsail Version:

2.5.1

### Browser:

Firefox

### Device:

Desktop PC

### Operating System:

Windows

### What happened?

Having multiple files in queue but I can't use it as a todo-list because I can only start the job added first from queue screen.

### What did you expect to happen instead?

Should be possible to start any file from queue not having to go look for it in files tab

### How to reproduce this bug?

add two file to queue
try to start second one

meteyou:
Why should this be a bug?

ristein:
Why should this be intentional?

th33xitus:
> Why should this be intentional?

It's currently the intended behavior. Should this be rather a feature request instead?

ristein:
that would be great

meteyou:
fixed with the PR #1692

PR #1692, opened 2023-12-16
feat: Add ability to re-arrange job queue's items

mdziekon:
## Description

This PR adds the ability to re-arrange job queue's items, by using the following actions:
- Move to the queue's start (make it the first item)
  - Displayed for all items beside first
- Move to the queue's end (make it the last item)
  - Displayed for all items beside last
- Move item up
  - Displayed for all items beside first & second
  - Second item already has "move to start", which is funtionally the same
- Move item down
  - Displayed for all items beside last & next to last
  - Next to last item already has "move to end", which is funtionally the same

All of the actions are available via the context menu (Right-click on the queue's entry). While "move up" & "move down" are probably not ideal for quick re-arrangements (drap & drop would be much easier to use), I think it's good enough for now, most definitely better than no such feature at all. The internal implementation is done is such a way, that if anyone would like to improve this by implementing the mentioned drag & drop, the re-arrangement "backend" is already there, ready to be reused. I think this could also be reused to implement the requested "Start printing from anywhere in the queue" feature, however I'm not versed enough in the Vue's ecosystem to implement that at this moment.

Since the modified component is properly being reused across Job Queue panel on the Dashboard and the Files list, the implemented features are available for both of these views.

Additional changes:
- `StatusPanelJobqueueEntry` component no longer relies of its status of being "first" to display the "print button"
  - This approach clashed with reusability of the implemented, new features
  - New approach (`showPrintButtonForFirst` flag) clearly represents the purpose, while the previous implementation required internal knowledge to understand why the button was being displayed or not

- (solves) https://github.com/mainsail-crew/mainsail/issues/1618
- (related) https://github.com/mainsail-crew/mainsail/issues/1379

(see next comment)

N/A


mdziekon:

#### One item only (no re-arrange) buttons

obraz

#### Multiple items, re-arrange first item

obraz

#### Multiple items, re-arrange second item

obraz

#### Multiple items, re-arrange middle item

obraz

#### Multiple items, re-arrange next to last item

obraz

#### Multiple items, re-arrange last item

obraz

#### Multiple items, job queue overflow (Dashboard only), re-arrange last DISPLAYED item (queue is longer)

obraz

#### Multiple items, Files view, re-arrange second to last item 

obraz

meteyou:
Thx for the PR. Would it be also possible with a "drag & drop" option like the resort function of "expert macro panel"?

mdziekon:
> Thx for the PR. Would it be also possible with a "drag & drop" option like the resort function of "expert macro panel"?

@meteyou

Hi, I'm not sure if you asked me to do that in this PR, or is it technically possible to achieve that in the future (as in, in a separate PR), so I'll answer both:
- Can I implement "drap & drop" in this PR?
    - No, currently I'm not experienced enough in the Vue ecosystem.
    - I did briefly experiment with `draggable` just now, and while I think the callback should supply necessary data to pass them into `changePosition` action, the styling gets completely wrecked. Looking at `draggable` usages, none uses tables to display the entries, so I'm not even sure if `draggable` is compatible with tables. But as I said, I'm not a "Vue guy" on a day to day basis, so maybe someone else has a better idea how to achieve it.
    - I think you'll need someone else to be a champion of said improvement. I would recommend treating this PR as a stop-gap solution, while waiting for someone else to step up and implement the drag & drop.
- Can "drag & drop" be achieved using currently implemented action?
    - Yes, I think so, as I said above while investigating `draggable` component. The `changePosition` action requires only the "final position", which as far as I can tell, `draggable` does provide in its callback.

meteyou:
Thank you for your reply. Please just let this PR open, and I'll take a closer look at it. Because the option "maintainers are allowed to edit this pull request", I can continue working on it.

Master92:
I'd just like to stop by and say thank you for this PR. This is exactly the feature I was just looking for at this very moment! 😃

Drag and drop support would be really nice, though. 👍

AdamBarnas:
I just started using job queue and first thing i thought was it needs this feature. So as soon as I started searching for it I found this PR. Great job @mdziekon. I'm looking forward to use it as soon as it is merged.

meteyou (approved):

PR #1714, opened 2023-12-25
fix: fix long initial time with huge print history

### LINKED ISSUE #1468: Slow website UI load times
### Mainsail Version:

v2.5.1 _(Edit: Just updated to v2.6.0 and still have the same issue)_

### Browser:

Chrome

### Device:

Desktop PC

### Operating System:

Windows

### What happened?

When I load the webpage it takes around 8-20 seconds to load. (Validated on multiple high-end devices). 

The page to load a lot faster than it currently does

If you have a lot of history file, you will see this slow down. I just with a lot of use time (438hours) 

### Additional information:

hso the issue seems to be that it loads the entire history on page I it. My suggestion is to have it load the top X from history (maybe 10?) and the rest of the history can be loaded on the history page load 

image

image

https://github.com/Rat-OS/RatOS-configuration/assets/4393979/61c7380c-8376-4ade-ab3c-a7b693c39aa6

https://github.com/Rat-OS/RatOS-configuration/assets/4393979/4aa5a27d-b621-418f-bfd6-3ccf810838ad

You can find my DB here: 
[data.zip](https://github.com/mainsail-crew/mainsail/files/12073713/data.zip)

meteyou:
## Description

This PR fix a issue with long initial times with huge print histories. It only loads the first 100 entries and the rest only when the user clicks on a button.

fixes #1468 

none

none


src/components/panels/HistoryListPanel.vue
```diff
@@ -523,14 +518,15 @@ import {
     mdiNotebookEdit,
     mdiNotebookPlus,
     mdiNotebook,
+    mdiDownload,
```
rackrick, line 521: remove unused icon reference

src/locales/de.json
```diff
@@ -351,6 +351,7 @@
         "Jobs": "Drucke",
         "LastModified": "Zuletzt geändert",
         "LayerHeight": "Schichthöhe",
+        "LoadCompleteHistory": "Lade komplette Verlaufsliste",
```
rackrick, line 354: We use `Historie` on different keys why the switch to `Verlaufsliste`?

```suggestion
        "LoadCompleteHistory": "Lade vollständige Historie",
```

src/locales/en.json
```diff
@@ -351,6 +351,7 @@
         "Jobs": "Jobs",
         "LastModified": "Last Modified",
         "LayerHeight": "Layer Height",
+        "LoadCompleteHistory": "Load complete history jobs",
```
rackrick, line 354:
```suggestion
        "LoadCompleteHistory": "Load complete history",
```

rackrick (approved):

jLynx:
Thanks for completing this, excited to try it out. Happy 2024!

jLynx:
Issue seems to be mostly solved now. Now it takes no longer than 8 seconds which is a massive improvement over 15-30 seconds previously. Looking at the natwork traffic, 2 seconds is for the css file. So I'm assuming thst gets dynamically generated?

meteyou:
@jLynx no... All files are static.

PR #1737, opened 2024-01-07
perf: batch gcode file metadata requests

nathanregner:
## Description

Batch `server.files.metadata` requests to Moonraker, resulting in fewer re-renders of the file table (once per batch, rather than once per file). The performance improvement is especially noticeable with lists of 100+ files.

Moonraker has supported JSON-RPC Batch Requests for [at least 4 years](https://github.com/Arksine/moonraker/blame/96e69240caa3bc6b5a7be93f88ea90a0e8e38035/moonraker/websockets.py#L109), so I don't think backwards incompatibility is a concern.

An alternative approach could be to optimize the GCode Files table itself, which currently re-renders in its entirety when file metadata is updated. However, this approach felt more straightforward (and I'm not super familiar with Vue).

### Before

https://github.com/mainsail-crew/mainsail/assets/9659564/a6c279bf-5d7f-416a-847a-55fca8b63325

before

### After

after

https://github.com/mainsail-crew/mainsail/assets/9659564/cc1188a9-a1c1-49d9-8cdc-896d4c6d7a65

### DIFF
--- a/src/components/panels/GcodefilesPanel.vue
+++ b/src/components/panels/GcodefilesPanel.vue
@@ -1235,11 +1235,12 @@ export default class GcodefilesPanel extends Mixins(BaseMixin, ControlMixin) {
 
     refreshMetadata(data: FileStateGcodefile[]) {
         const items = data.filter((file) => !file.isDirectory && !file.metadataRequested && !file.metadataPulled)
-        items.forEach((file: FileStateGcodefile) => {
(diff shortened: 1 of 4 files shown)


meteyou (approved):

meteyou:
@nathanregner Thanks for this awesome PR. I don't know why I overlooked this function in Moonraker until now.

PR #1749, opened 2024-01-19
refactor: refactor spoolman integration to support v2 response

meteyou:
## Description

This PR adds the support of the V2 response from Moonraker.

PR for the V2 response in Moonraker: https://github.com/Arksine/moonraker/pull/792

none

none

### DIFF
--- a/src/store/server/spoolman/actions.ts
+++ b/src/store/server/spoolman/actions.ts
@@ -3,6 +3,18 @@ import { ActionTree } from 'vuex'
 import { RootState } from '@/store/types'
 import { ServerSpoolmanState } from '@/store/server/spoolman/types'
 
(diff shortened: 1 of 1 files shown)


src/store/server/spoolman/actions.ts
```diff
@@ -15,6 +27,7 @@ export const actions: ActionTree<ServerSpoolmanState, RootState> = {
             {
                 request_method: 'GET',
                 path: '/v1/info',
+                use_v2_response: true,
```
rackrick, line 30: this and the following uses of `use_v2_response` require a minimum moonraker right?

meteyou: this is backward compatible. if Moonraker is older, this value will be ignored and the answer will also be converted.

rackrick (approved):

PR #1777, opened 2024-02-14
fix: file upload rate displays `NaN` instead of an actual value

mdziekon:
## Description

This PR fixes incorrect upload (and download in GCode viewer, although barely noticeable) rate being displayed as `NaN`, instead of an actual value.

Problem is caused by an incorrectly used `progressEvent` argument of Axios'es `onUploadProgress` callback, which exposes its own event object wrapper, instead the raw browser event (as I think someone has assumed, or maybe was the case in the past). Now, with Axios 1.x being used, `onUploadProgress` callback exposes `AxiosProgressEvent`-typed interface (this PR also corrects typings usage, instead of the nasty implicit `any`).

[`AxiosProgressEvent` "documentation"](https://github.com/axios/axios/blob/v1.6.0/index.d.ts#L287-L297) for reference (typings, their official documentation does not seem to describe the relevant interface).

Problem occurs on Firefox & Chromium based browsers.

Problem has been fixed for:
- File explorer upload
- "Upload & Print" topnav functionality
- GCode viewer ("Current file" download snackbar)

N/A

_Note: upload artificially slowed down for testing purposes_

### Before

obraz

### After

obraz

### DIFF
--- a/src/components/TheTopbar.vue
+++ b/src/components/TheTopbar.vue
@@ -110,7 +110,7 @@ import { Mixins } from 'vue-property-decorator'
 import BaseMixin from '@/components/mixins/base'
 import { validGcodeExtensions } from '@/store/variables'
 import Component from 'vue-class-component'
-import axios from 'axios'
+import axios, { AxiosProgressEvent } from 'axios'
 import { formatFilesize } from '@/plugins/helpers'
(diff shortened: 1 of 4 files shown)


meteyou (approved):

meteyou:
Thx for this fix! I didn't even notice this bug until just now. Looks like i overlooked an Axios update.

PR #1274, opened 2023-02-18
feat: Reminders panel on the History page

jeremiahvpratt:
### This PR adds a Reminders panel, which allows users to create and manage incremental reminders based on an elapsed number of print hours since reminder creation.
This PR will fix #1129 

### How to use it:

1. Ensure history is enabled in your moonraker conf
2. In the History tab, click the "+" button in the Reminders panel
3. Add a name and print hours duration to your reminder
4. Print for that many hours, then observe a reminder appearing in notifications.
5. Repeat a reminder by right clicking it and selecting "repeat" - this resets the start time for the reminder to the current number of total print hours on the machine.
6. Delete a reminder by right clicking it and selecting "delete", or selecting multiple reminders with their checkboxes and clicking the red trash icon at the top of the Reminders panel.
7. Edit a reminder by right clicking it and selecting "edit".

### What's missing:
Translations, I'm not sure what the process is for adding them.

### Images:
**Reminders Panel**
Panel
**Create Dialog**
create
**Notification**
notification

Signed off by: Jeremiah Pratt <jeremiah.v.pratt@gmail.com>

PR #1778, opened 2024-02-14
feat: add ability to add history items to job queue

mdziekon:
## Description

This feature request & PR implements new context menu entry in History view, allowing users to add history entries to Queue. Right now, Mainsail only allows to start reprinting historical items immediately, which in certain cases might not be desired (eg. when user wants to print that file later; user is already printing something but wants to quickly schedule next prints from the History view; user wants to reprint a file, but with a different Spool which has to be changed via Spoolman panel on the Dasboard).

After adding an item to the Queue, a toast is being presented, indicating that something has happened. This was added mostly because adding to queue does not force any screen change (and it shouldn't, as there's no reason to do that, users might want to continue adding other historical entries to the Queue), so there should be some form of feedback.

~This PR adds only the "Add to Queue" menu item.~

This PR:
- adds both "Add to Queue" & "Add batch to Queue" menu items.
- refactors and componentizes "Add batch to Queue" dialog used in File explorer, Status files list & (now) History entries list, making it reusable and self-contained.
- fixes form validation - previously, even though `count` field validation was applied and displayed "on input", it would not prevent users from submitting the form, by either pressing "Enter" key on the field, or clicking on the "Add to Queue" button.

~"Add batch to Queue" is possible, however would require a bit of refactoring (the "Add batch" modal should be refactored into a separate component, to make it reusable and prevent yet another, third I think, implementation to exist). I might take this up as well, but since the basic functionality already works, I believe this can be picked up in the future (in case if I won't be able to pick this up in this PR - let's not stall feature delivery if it's possible to take the iterative approach).~

N/A

obraz
obraz

### DIFF
--- a/src/components/dialogs/AddBatchToQueueDialog.vue
+++ b/src/components/dialogs/AddBatchToQueueDialog.vue
@@ -67,6 +67,7 @@ export default class AddBatchToQueueDialog extends Mixins(BaseMixin) {
      * Should there be a toast message after the file was added to the queue?
      */
     @Prop({ type: Boolean, default: false }) declare readonly showToast: boolean
+
     /**
      * Filename of the model to be added to the Queue.
      */
--- a/src/components/dialogs/AddBatchToQueueDialog.vue
+++ b/src/components/dialogs/AddBatchToQueueDialog.vue
@@ -74,7 +74,8 @@ export default class AddBatchToQueueDialog extends Mixins(BaseMixin) {
     @Prop({ type: String, required: true }) declare readonly filename: string
 
     isValid = false
-    input: number = 1
+    // because of the text field, the input is always a string
+    input: string = '1'
 
     rules = {
         count: [
@@ -84,7 +85,7 @@ export default class AddBatchToQueueDialog extends Mixins(BaseMixin) {
     }
 
     async addBatchToQueueAction() {
-        const array: string[] = new Array(this.input).fill(this.filename)
+        const array = Array(parseInt(this.input)).fill(this.filename)
 
         await this.$store.dispatch('server/jobQueue/addToQueue', array)
 
@@ -99,7 +100,7 @@ export default class AddBatchToQueueDialog extends Mixins(BaseMixin) {
     }
 
     resetFormState() {
-        this.input = 1
+        this.input = '1'
     }
 
     @Watch('isVisible')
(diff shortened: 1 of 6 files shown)


meteyou (approved):

meteyou:
@mdziekon thank you very much! I just did some refactoring myself, because we are short before the next release and i did here (https://github.com/mainsail-crew/mainsail/pull/1274) a complete refactoring of the history page.

If you have some questions, why i did some things in your code, feel free to ask here!

src/components/dialogs/AddBatchToQueueDialog.vue
```diff
@@ -0,0 +1,120 @@
...
+
+    rules = {
+        count: [
+            (value: string) => !!value || this.$t('JobQueue.InvalidCountEmpty'),
+            (value: string) => parseInt(value, 10) > 0 || this.$t('JobQueue.InvalidCountGreaterZero'),
+        ],
+    }
+
+    async addBatchToQueueAction() {
+        const array = Array(parseInt(this.input)).fill(this.filename)
```
mdziekon, line 88: @meteyou 

Note about the `parseInt` usage - in my unmodified code, I've added the `radix` argument on purpose, to force the "natural" parsing of inputs, as in "treat strings as natural number input". Typing something like "0x11" makes `parseInt` assume it's a hexadecimal, and while it is most likely a super rare case, I usually tend to prevent ambiguity in terms of parsing, that's why I've included it. Not sure if you know about this or not, and I also don't mind this being gone here (as I said, rather rare to see that kind of typing mistake made by the users), but just wanted to clarify in case you didn't know about this.

While this does not need any action, please note that the validation rule above still leverages the `radix = 10` argument, so the "wrong" (or rather, unnatural) input is impossible to be submitted. However, I think it's usually good to enforce same style of this function's usage, so if you agree with my take, I would recommend adding the radix back in some future refactoring task. This can also be enforced (autofixed even, I think) with ESLint rule: https://eslint.org/docs/latest/rules/radix (`always` option required, because `as-needed` incorrectly assumes that `10` is the default).

meteyou: ohh thx! i didn't know that! Do you want to add this eslint rule with a PR? I think this would be a good think to "fix" this issues in the complete project.

mdziekon: Sure, I can pick this up, shouldn't be much of a problem :)

PR #1800, opened 2024-03-02
fix: fix long M117 outputs in the status panel

meteyou:
## Description

This PR fix long M117 outputs in the status panel.

none

before:
image

after:
image

none


rackrick (approved):
error only reproducible with status text without any spaces. who use such texts? oO

meteyou:
Long gcode filenames also create this issue, because klipper posts it at the print start. I don't think that this happens with normal macro outputs.

here is an example:
no_name

PR #1829, opened 2024-03-21
fix: ignore wrong default.json file while resetting moonraker db

meteyou:
## Description

This PR will ignore a wrong/not parseable default.json file during a moonraker db reset.

none

none

none

### DIFF
--- a/src/components/settings/General/GeneralReset.vue
+++ b/src/components/settings/General/GeneralReset.vue
@@ -86,7 +86,7 @@ export default class SettingsGeneralTabResetDatabase extends Mixins(BaseMixin, S
         this.resetableNamespaces = await this.loadBackupableNamespaces()
 
         // stop if history is not enabled
-        if (this.moonrakerComponents.includes('history')) return
+        if (!this.moonrakerComponents.includes('history')) return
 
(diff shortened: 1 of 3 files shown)


src/components/settings/General/GeneralReset.vue
```diff
@@ -85,17 +85,18 @@ export default class SettingsGeneralTabResetDatabase extends Mixins(BaseMixin, S
     async loadResetableNamespaces() {
         this.resetableNamespaces = await this.loadBackupableNamespaces()
 
-        if (this.moonrakerComponents.includes('history')) {
-            this.resetableNamespaces.push({
-                value: 'history_jobs',
-                label: this.$t('Settings.GeneralTab.DbHistoryJobs'),
-            })
+        // stop if history is not enabled
+        if (this.moonrakerComponents.includes('history')) return
```
rackrick, line 89: shouldn't that be an inverted check?
the comment says `is not enabled` but you check for enabled history

meteyou: Thanks! You're absolutely right,

dw-0 (approved):

PR #1830, opened 2024-03-22
feat: Disable Zoom on Mobile

Helvio88:
The reason zooming should be disabled, is because double tapping buttons (e.g. moving an axis) causes zooming.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
     <head>
         <meta charset="utf-8" />
         <meta http-equiv="X-UA-Compatible" content="IE=edge" />
-        <meta name="viewport" content="width=device-width,initial-scale=1.0" />
+        <meta name="viewport" content="width=device-width,initial-scale=1.0, user-scalable=no" />
         <title>Mainsail</title>
         <meta name="description" content="Mainsail is the popular web interface for Klipper." />
 


meteyou:
Hey! Thx for your PR, but i cannot reproduce this issue on my phone. Which OS and browser do you use?

I think to disable scaleable ist the wrong way, because there are also some tablet users, which use the scale function.

Helvio88:
Hey there! I'm using Safari (Web App) on iPhone. I sometimes try to baby-step the Z axis using the buttons and the double-tapping zooms the page, which is undesirable.

Alternative solution (Global CSS):
```css
  * {
      touch-action: manipulation;
  }
```

Docs: https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action#manipulation

meteyou:
can you please specify the selector, that it only operates for buttons?

Helvio88:
I could, it would be
`button input[type="button"]`

But you and I have different opinions on how wide this setting should be.

I would suggest this PR is closed and you accept the idea as a suggestion for improvement, and implement however it makes more sense to you.

Thank you for considering!

meteyou:
i just don't want to "break" anyone else's workflow, but i understand your issue and am trying to fix it as best as possible.

Issue #1599, opened 2023-10-10
Add support for moonrakers at paths


Laikulo:
### Requested feature:

Add support for moonraker instances served at paths other than /, either with moonraker's new `route_prefix` option, or with reverse proxy settings.

### Solves the following problem:

Allows connecting to multiple moonrakers served over the same port, allowing for a host running multiple printers to do so in a more elegant way.

meteyou:
How should this work? You have to connect to Moonraker, to read the instances?

meteyou:
This FR was fixed with the PR #1836

PR #1836, opened 2024-03-30
feat: Connect to Moonraker via subdirectory/path

4cello:
## Description

This allows to connect to moonraker instances on non-root paths, e.g. `all-printers.net/myprinter`.

This was already possible by appending the path to the port: `7125/myprinter`, which is hacky and non-intuitive.

I added an optional field "path" to the printers. The text fields are hidden in subpanels in `SelectPrinterDialog` and `SettingsPrinterRemoteTab`. The only input validation currently is to check that the path starts with a "/", which is in line with validation rules for the hostname.

### Things I don't know if they're necessary or how to do (yet)
- Modify the default webcam path to respect this change: I'm not sure what the most elegant way to do this would be.
- Moonraker instance database: I don't know if there's any work to do here

Addresses #1599 as well as @PSandro's comment on #1163 

mainsail-printersettings-2024-03-30_11-03
mainsail-editprinter-2024-03-30_11-03


meteyou:
Thanks for your PR!

Can you only remove the "advanced" accordion in the interface settings dialog and just display path per default with the default value "/"?

And pls use the full width in the small "select printer" dialog for the path input field (when no name field is visible. so when moonraker is used for the instance db)

4cello:
Will do!

What are your thoughts on the accordion in the small dialog? I don't love the look of it in general, but I also think it's good to hide an option which most people don't want to change anyways.

meteyou:
to be honest, i personally don't like this function at all...

maybe you could add a button at the bottom left next to the delete button to show/hide the advanced options?

4cello:
I think it could work, but I'm struggling to find a good icon. The good ol' Expand/Collapse just doesn't feel right when it's below the thing that it controls, and I also think it looks better if there's no text required.

mainsail-collapsed-2024-04-01_01-05

mainsail-expanded-2024-04-01_01-06

mainsail-gear-2024-04-01_01-16

meteyou:
ok. i think i found an issue, when you try to connect to a printer without sub path:
image

meteyou (approved):

meteyou (approved):

Issue #1731, opened 2024-01-04
Direct link to specific printer in multi-printer setup


kondas:
### Requested feature:

The possibility to  have a direct link to specific printer when multiple printers are handled by a single Mainsail instance

### Solves the following problem:

OrcaSlicer can send print jobs directly ro Moonraker API, but to have device control only an embedded web browser is available. 
It would be nice to have direct link to a specific printer in Mainsail for this purpose, because currently the printer has to be selected after switching to device tab.

### Additional information:

Happy new year!

meteyou:
thanks for your request. something like this has been asked for several times, but i don't think it will be able with every variant. it might be most useable with the json instances store. how have you currently configured your multi instances?

kondas:
I'm using the json store. Is it possible already? I've tried searching to no avail

kondas:
Thanks for clarification. 

If there is possibility to modify something in the HTTP request from the client would be adequate, that way one could make each printer accessible under a separate hostname by ie. setting a header value in the reverse proxy

napter:
Is there still no way to add param to the url so mainsail starts with that printer? It would be a lot cleaner if we could create a bookmark to link to a specific printers dashboard.

meteyou:
@napter this is an open feature request.. so no. there is no integrated function for that.

meteyou:
this FR is fixed with the PR #1837

PR #1837, opened 2024-03-30
feat: Direct link to specific printer via query parameter

4cello:
## Description
Adds the ability to directly connect to a printer via an URL query parameter, e.g. `my-mainsail.net/cam?printer=myprinter`.

To accomplish this, added an optional field `name` to the printers (in browser/json mode). Whenever the `SelectPrinterDialog` is mounted, it checks the URL params and connects to the printer with the matching name, if any. Otherwise the selection dialog is displayed without an error.

I added the text field for the name to the `SelectPrinterDialog` and `SettingsPrinterRemoteTab` in an expandable sub-panel. There is currently no input validation, or checks for duplicate names. 

If one were to implement a duplicate check, the name could instead be used in place of the ID, so a printer ID would either be a user specified string or a UUID. I don't have opinions either way.

### TODO
- [x] What is the Moonraker database, and what would need to be changed? (see my other PR #1836)

Simple implementation for #1731

mainsail-addprinter-name-2024-03-30_11-24
Peek 2024-03-30 11-31


meteyou:
just a short feedback here, i would not include this feature for moonrakerDB instances... i think this is not "useful/useable" in this case. if a user will use this, should change the config.json for example.

4cello:
I assumed that would be the case. 
The current implementation just filters the printer list for the given name, and if there's none or multiple printers, it just shows the selection dialog (and prints the error to the console), so it should not be invasive at all.

meteyou (approved):

Issue #1879, opened 2024-05-10
404 Moonraker request without Maintenance


mryel00:
### Mainsail Version:

v2.11.1

### Browser:

Chrome

### Device:

Desktop PC

### Operating System:

Windows

### What happened?

If someone has no Maintenance created, there will be always a request on startup to Moonraker to a path that doesn't exist:
`[application.py:log_request()] - 404 GET /server/files/config/.theme/maintenance.json?time=1715354643052`

### What did you expect to happen instead?

According to @meteyou this should only happen the first time opening Mainsail with no entry.

### How to reproduce this bug?

Have no Maintenance entry in your history and reload Mainsail once.

PR #1914, opened 2024-06-16
fix(maintenance): add init store entry to init maintenance store only one time

meteyou:
## Description

This PR fixes the init maintenance store routine. It will add one "MAINTENANCE_INIT" entry in the Moonraker DB.

fixes: #1879 

none

none

### DIFF
--- a/src/store/gui/maintenance/actions.ts
+++ b/src/store/gui/maintenance/actions.ts
@@ -36,6 +36,14 @@ export const actions: ActionTree<GuiMaintenanceState, RootState> = {
         // stop, when no entries are available/found
         const entries = defaults.entries ?? []
         if (entries?.length === 0) {
+            Vue.$socket.emit('server.database.post_item', {
+                namespace: 'maintenance',
+                key: uuidv4(),
+                value: {
+                    name: 'MAINTENANCE_INIT',
+                },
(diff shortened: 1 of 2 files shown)


src/store/gui/maintenance/actions.ts
```diff
@@ -92,7 +100,12 @@ export const actions: ActionTree<GuiMaintenanceState, RootState> = {
 
     async initStore({ commit, dispatch }, payload) {
         await commit('reset')
-        await commit('initStore', payload)
+
+        const entries = payload.value ?? {}
+        const initKey = Object.keys(entries).find((key) => entries[key]?.name === 'MAINTENANCE_INIT')
+        if (initKey) delete entries[initKey]
```
rackrick, line 106: `delete` sets the `MAINTENANCE_INIT`entry to `undefined`. is this intended?
to remove the entry the `splice()` function would be better

meteyou: entries is an object and not an array. i think splice only works on arrays.

rackrick: Ohh... you're right. Stupid.

rackrick (approved):

PR #1948, opened 2024-07-16
feat: Alias name some printer objects / macro 

CF3B5:
## Description
Alias name some printer objects / macro, so that other languages look nice :-)

none

Setting
Setting Alias name
Setting Macro alias name

Dashboard panel
Temperatures
Macro
Miscellaneous


meteyou:
Many thanks for this PR, but unfortunately, I cannot agree with it. From the beginning of Mainsail, we have had the principle that all names in Mainsail correspond 100% to the names from the printer config, and we want to keep this behavior. I have discussed this internally with the whole crew again, and we want to keep this principle.

Can your language be used for Klipper objects with names? As an example:

```
[gcode_macro hotend_pos]
```
Can you convert the name "hotend_pos" here to your language?

If this were possible, only objects where the module name is also the object name would remain, such as `[extruder]`, `[heater_bed]`, etc. Then, the question is whether we can also include the possibility of translating these with i18n (optional).

Just my two cents on the general topic of "translations": My mother language is not English, but I prefer English, especially in the 3D printing business. All the terms you read on the internet when you need help or when you work on the config are in English. You can't get far without English, especially since you can't confuse terms by mistranslations.

PR #1953, opened 2024-07-20
fix(control): check set actionButton before display it

meteyou:
## Description

This PR fix the action button in the Cross & Circle control. When you set the action button to QGL and then remove QGL, then it will display the default action button instead of QGL now.

none

none

none


src/components/panels/ToolheadControls/CrossControl.vue
```diff
@@ -377,7 +377,16 @@ export default class CrossControl extends Mixins(BaseMixin, ControlMixin) {
...
+        const button = this.$store.state.gui.control.actionButton ?? this.defaultActionButton
+
+        if (button === 'qgl' && !this.$store.getters['printer/existsQGL']) {
+            return this.defaultActionButton
+        }
+        if (button === 'ztilt' && !this.$store.getters['printer/existsZTilt']) {
+            return this.defaultActionButton
+        }
+
+        return button
```
rackrick, line 389: duplicate code... could be moved into the control mixin

src/components/panels/ToolheadControls/CrossControl.vue
```diff
@@ -377,7 +377,16 @@ export default class CrossControl extends Mixins(BaseMixin, ControlMixin) {
     mdiHome = mdiHome
 
     get actionButton(): string {
-        return this.$store.state.gui.control.actionButton ?? this.defaultActionButton
+        const button = this.$store.state.gui.control.actionButton ?? this.defaultActionButton
+
+        if (
```
rackrick, line 382: combining the two ifs doesn't change the fact the same logic is used in both Controls `CrossControl.vue` and `CircleControl.vue` :D
But fine we can leave it like this

rackrick (approved):

rackrick (approved):

PR #1956, opened 2024-07-21
fix(notifications): fix dismiss function for tmc warnings

### LINKED ISSUE #1955: Driver OTPW cannot be cleared or ignored
### Mainsail Version:

2.12.0

### Browser:

Chrome

### Device:

Desktop PC

### Operating System:

Windows

### What happened?

A lot of buttons that do nothing. This seems unintentional. 

https://github.com/user-attachments/assets/aa1bc50c-e689-40c8-acd8-275cacc8ab90

It would be nice if they worked. Currently you cannot avoid this warning if you want to get the maximum out of a set of 2209 sticks for example, and as such this is just an annoyance and not very helpful.

The notification to be dismissed / ignored, or worst case scenario that the buttons were disabled.

Either generate the OTPW flag on a driver, or just unconditionally push it to the notification list in the client code since it's a mainsail-only implementation.

### Additional information:

_No response_

meteyou:
## Description

This PR fix the dismiss function for TMC warnings in the notifications.

fixes #1955 

none

none


src/store/gui/notifications/getters.ts
```diff
@@ -436,7 +436,17 @@ export const getters: GetterTree<GuiNotificationState, any> = {
...
+        // get all dismissed tmcwarnings and convert it to a string[]
+        const tmcwarningsDismisses = getters['getDismissByCategory']('tmcwarning').map(
+            (dismiss: GuiNotificationStateDismissEntry) => {
+                return dismiss.id
+            }
+        )
+
+        // return filtered tmcwarnings
+        return notifications.filter((entry) => {
+            return !tmcwarningsDismisses.includes(entry.id.substring(11))
```
rackrick, line 448: magic number... what is part of the string which starts at character 11?

rackrick (approved):

PR #1978, opened 2024-08-17
feat: added second layer confirmation for Cancel Job

FatalBulletHit:
## Description

This PR adds a second layer confirmation dialog for the Cancel Job button.

None.

No visual changes (same confirmation dialog as emergency stop).

I have no idea about Docker and the likes, so I haven't been able to test this yet.
Translation missing for all languages but English and German.

### DIFF
--- a/src/components/dialogs/CancelJobDialog.vue
+++ b/src/components/dialogs/CancelJobDialog.vue
@@ -1,0 +1,50 @@
+<template>
+    <v-dialog :value="showDialog" width="400" persistent>
+        <panel
+            :title="$t('CancelJobDialog.CancelJob')"
+            toolbar-color="error"
+            card-class="cancel-job-dialog"
+            :icon="mdiStop"
+            :margin-bottom="false">
+            <template #buttons>
+                <v-btn icon tile @click="closePrompt">
+                    <v-icon>{{ mdiCloseThick }}</v-icon>
+                </v-btn>
+            </template>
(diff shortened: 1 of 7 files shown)


meteyou:
Thx for your PR!

its fine, that you only add languages, which you are know/speak. all other languages should be translated by the translators. (missing translations are better then wrong translations, because the translators will find them faster).

you don't need docker for testing it. you just have to:
- install npm on your pc
- run `npm ci`
- rename and modify the .env file
- add `http://localhost:8080` to the `cors_domains` in your `moonraker.conf` of your printer
- run Mainsail in dev mode `npm run serve`
- open Mainsail with `http://localhost:8080`

then you should be able to test it.

please change the toolbar from red to normal. the e-stop dialog is the only one that has this toolbar color. every other confirm dialog does not have this.

furthermore, i am still thinking about how useful this dialog is, because it is actually a third level confirm (pause -> cancel -> dialog). but it is optional, and there are users who deactivate the pause -> cancel, so it could be useful especially for these users.

FatalBulletHit:
Fixed, tested and working, thanks for the help!
Will use this with the "always display cancel print" option and hopefully not accidentally cancel prints anymore. 😄
image

meteyou:
pls also change the icon from the dialog. this one looks like a "loading issue" from the icon.

FatalBulletHit:
How about `mdiStopCircleOutline`? Or would you prefer another one or no icon at all?
image

meteyou (approved):

PR #1988, opened 2024-09-01
refactor(ControlPanel): use SAVE/RESTORE STATE when moving

meteyou:
## Description

This PR should switch to SAVE_GCODE_STATE/RESTORE_GCODE_STATE to set everytime back to the original movement mode.

Comment from @pedrolamas in the extrude panel PR: https://github.com/mainsail-crew/mainsail/pull/1965#issuecomment-2322985433

https://github.com/user-attachments/assets/7bfa804c-8589-4467-84ae-2a6b7a18ed31

none


zellneralex:
Tested several edge cases:
- very slow moves (1mm/sec) and tried to modify other values like babystepping, speed and extruder factor
- multiple hitting of the move arrow while slow and fast (100 mm/sec) moves. System starts in absolute and must also end in that mode.

All tests where PASS so that I would approve the change

pedrolamas:
I guess the question now is in terms of value provided: I do think this makes sense, but does it solve a problem or provide new value?

The only downside I could see is that it adds a lot of "noise" to the logs... but even that is mitigatable, I think!

meteyou:
> I guess the question now is in terms of value provided: I do think this makes sense, but does it solve a problem or provide new value?
> 
> The only downside I could see is that it adds a lot of "noise" to the logs... but even that is mitigatable, I think!

It only brings the "added value" that the previous mode is always set again. Some users, for example, do not have absolute mode in their start macro. We have already had a few support-requests as to "why the printer cannot print after a movement" (out-of-range message).

Where you should say, “Please fix your gcode/macros”... So it's a trade-off between “error protection” vs “noise in the console”.

src/components/mixins/control.ts
```diff
@@ -135,9 +135,11 @@ export default class ControlMixin extends Vue {
     }
 
     doSendMove(gcode: string, feedrate: number) {
-        gcode = 'G91' + '\n' + 'G1 ' + gcode + ' F' + feedrate * 60
-
-        if (this.absolute_coordinates) gcode += '\nG90'
+        gcode =
+            `SAVE_GCODE_STATE NAME=ui_movement\n` +
```
pedrolamas, line 139: I know there is no value in this change, but keeping consistency, I would argue that the name should be prefixed with "_" to mark it as "internal".

```suggestion
            `SAVE_GCODE_STATE NAME=_ui_movement\n` +
```

FWIW, Klipper does exactly this internally on [firmware_retraction](https://github.com/Klipper3d/klipper/blob/08a1c9f12760ee6d89db2b82e76c7d93453212db/klippy/extras/firmware_retraction.py#L56) module.

src/components/mixins/control.ts
```diff
@@ -135,9 +135,11 @@ export default class ControlMixin extends Vue {
...
+            `G91\n` +
+            `G1 ${gcode} F${feedrate * 60}\n` +
+            `RESTORE_GCODE_STATE NAME=ui_movement`
```
pedrolamas, line 142:
```suggestion
            `RESTORE_GCODE_STATE NAME=_ui_movement`
```

meteyou:
@pedrolamas thx for your suggestion! you are complete right. i also opened a PR to rename the gcode_state name in the control panel.

rackrick (approved):

PR #2004, opened 2024-09-09
chore(websocket): add function to send and wait for response

meteyou:
## Description

This PR only implement a function to the WebsocketClient to send a request and wait for the response. Thx @Spice-King for this implementation/syntax.

none

none

none

### DIFF
--- a/src/plugins/webSocketClient.ts
+++ b/src/plugins/webSocketClient.ts
@@ -29,6 +29,13 @@ export class WebSocketClient {
     handleMessage(data: any) {
         const wait = this.getWaitById(data.id)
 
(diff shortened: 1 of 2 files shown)


src/plugins/webSocketClient.ts
```diff
@@ -68,8 +70,13 @@ export class WebSocketClient {
             return
         }
 
+        // resolve promise if it exists
+        if (wait?.resolve && 'result' in data) {
```
rackrick, line 74: we check if the wait exists a couple of lines before. So there no need for the null check operator.

rackrick (approved):

PR #2006, opened 2024-09-10
feat(Heightmap): add option to set the default orientation

### LINKED ISSUE #993: Improve default perspective and readability of heightmap graph
### Requested feature:

This is a feature request that only makes sense if everyone agrees (as most of them).. 
on the mesh view, the default view it's with the Y0 XMax closer to you.. that's a really weird prespective I think, it's hard for me to understand where's what, I have to move the thing arround until I get a 0,0 visible.. My proposal is to have it from the front of the printer, like I phisically look at it.. 
it would also be nice to default to a zoom where everything is in view, including labels.. right now I need to zoom out so I can see the axis labels.
Maybe even the panel height can be increased. 
Default view: 
image
Preferred view: 
image

_No response_

### Additional information:

_No response_

meteyou:
## Description

This PR adds an option in the Settings to let the user set the default orientation of the bed mesh visualization.

fixes #993 

Option "Right Front" (default option and the same orientation as before):
image

Option "Left Front":
image

Option "Front":
image

Option "Top":
image

none

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -1021,7 +1021,7 @@
         "HeightmapTab": {
             "ColorSchemes": "Color Schemes",
             "DefaultOrientation": "Default Orientation",
-            "DefaultOrientationDescription": "Select the default view for the bed mesh visualization.",
+            "DefaultOrientationDescription": "Select the default orientation for the bed mesh visualization.",
             "Heightmap": "Heightmap",
             "IsDefault": "(Default)",
             "Orientations": {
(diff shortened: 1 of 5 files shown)


src/locales/en.json
```diff
@@ -1020,8 +1020,16 @@
         },
         "HeightmapTab": {
             "ColorSchemes": "Color Schemes",
+            "DefaultOrientation": "Default Orientation",
+            "DefaultOrientationDescription": "Select the default view for the bed mesh visualization.",
```
rackrick, line 1024:
```suggestion
            "DefaultOrientationDescription": "Select the default orientation for the bed mesh visualization.",
```

rackrick (dismissed):

rackrick (approved):

Issue #2009, opened 2024-09-13
File thumbnails broken in History Panel when file is placed inside folder(s)


mdziekon:
### Mainsail Version:

2.12.0

### Browser:

Firefox

### Device:

Desktop PC

### Operating System:

Linux

### What happened?

When using History page / panel, files which are placed in folder(s) do not have proper thumbnail displayed, making it much harder to quickly find entries we are looking for. Files placed in the root directory do not exhibit this behavior.

![thumbnail_bug](https://github.com/user-attachments/assets/73673dc0-957c-4209-bdc7-4c77a09d3010)

### What did you expect to happen instead?

Files with thumbnails available should have that thumbnail displayed, regardless of their location in file system.

### How to reproduce this bug?

1. Create a folder
2. Place a model (with a thumbnail) in that folder
3. Print it (or at least attempt to print it so that it appears in History page)
4. Go to History page

### Additional information:

This appears to be a regression caused by the refactoring done in #1274. The culprit seems to be incorrect creation of thumbnail url for "small thumbnails", which incorrectly removes the final `/` character needed to build correct relative thumbnail path. This does not occur for "big thumbnails".

Previous, working code:
https://github.com/mainsail-crew/mainsail/pull/1274/files#diff-21c7ea907f2434c89b31411768fd6e832d8b9afd3597a8a5cdb835343ef03319L903-L944

New, broken code:
https://github.com/mainsail-crew/mainsail/pull/1274/files#diff-26b52f4bf48f9c0010205b6d9b658c3f1f329ceaea65dae5bfc28d7b62058426R150-R180

I'll post a patch for this bug shortly, as I've already fixed it in my local version.

meteyou:
@mdziekon your links to the preview/new code doesn't work for me. can you pls a link to the file with line itself and not in the PR view?

mdziekon:
@meteyou right, that's probably because Github collapses both of these files by default, therefore hotlinking to specific lines does not work as intended.

Here's perma-link to working code:
https://github.com/mainsail-crew/mainsail/blob/8e7db0da04915f734fa8802ae76d5f283621f4de/src/components/panels/HistoryListPanel.vue#L903-L944

Here's perma-link to broken code:
https://github.com/mainsail-crew/mainsail/blob/866a0c6a49d90bb739d3b8c451163c9f196a0540/src/components/panels/History/HistoryListRow.vue#L150-L180

PR #2010, opened 2024-09-13
fix(HistoryPanel): Fix History thumbnails of files in folders

mdziekon:
## Description

Fixes missing thumbnails in History page for files placed in non-root folders.

Bonus: refactors thumbnails url generation code so that it is reused for both small & big thumbnails, and improves TypeScript coverage by eliminating unwanted `any` usage.

#2009

### Before

obraz

### After

obraz

### DIFF
--- a/src/components/panels/History/HistoryListEntryJob.vue
+++ b/src/components/panels/History/HistoryListEntryJob.vue
@@ -195,7 +195,7 @@ export default class HistoryListPanel extends Mixins(BaseMixin) {
                 thumb.height <= thumbnailSmallMax
         )
 
-        return this.createThumbnailUrl(thumbnail)
+        return thumbnail ? this.createThumbnailUrl(thumbnail) : false
     }
 
     get bigThumbnail() {
@@ -203,7 +203,7 @@ export default class HistoryListPanel extends Mixins(BaseMixin) {
 
         const thumbnail = this.item.metadata?.thumbnails?.find((thumb) => thumb.width >= thumbnailBigMin)
 
-        return this.createThumbnailUrl(thumbnail)
+        return thumbnail ? this.createThumbnailUrl(thumbnail) : false
     }
 
     get statusIcon() {
@@ -315,7 +315,7 @@ export default class HistoryListPanel extends Mixins(BaseMixin) {
         }
     }
 
-    createThumbnailUrl(thumbnail: FileStateFileThumbnail | undefined) {
+    createThumbnailUrl(thumbnail: FileStateFileThumbnail) {
         let relative_url = ''
         if (this.item.filename.lastIndexOf('/') !== -1) {
             relative_url = this.item.filename.substring(0, this.item.filename.lastIndexOf('/') + 1)
--- a/src/components/panels/History/HistoryListEntryJob.vue
+++ b/src/components/panels/History/HistoryListEntryJob.vue
@@ -321,9 +321,7 @@ export default class HistoryListPanel extends Mixins(BaseMixin) {
             relative_url = this.item.filename.substring(0, this.item.filename.lastIndexOf('/') + 1)
         }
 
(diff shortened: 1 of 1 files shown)


src/components/panels/History/HistoryListEntryJob.vue
```diff
@@ -187,40 +188,22 @@ export default class HistoryListPanel extends Mixins(BaseMixin) {
...
-            relative_url = this.item.filename.substring(0, this.item.filename.lastIndexOf('/') + 1)
-        }
-
-        if ((thumbnail?.relative_path ?? null) === null) return false
+        const thumbnail = this.item.metadata?.thumbnails?.find((thumb) => thumb.width >= thumbnailBigMin)
 
-        return `${this.apiUrl}/server/files/gcodes/${encodeURI(relative_url + thumbnail?.relative_path)}?timestamp=${
-            this.item.metadata.modified
-        }`
+        return this.createThumbnailUrl(thumbnail)
```
meteyou, line 206:
```suggestion
        return thumbnail ? this.createThumbnailUrl(thumbnail) : false
```

or 

```suggestion
        if (thumbnail === undefined) return false
        
        return this.createThumbnailUrl(thumbnail)
```

i would not enter the createThumbnailUrl function, when the thumbnail was not found.

src/components/panels/History/HistoryListEntryJob.vue
```diff
@@ -331,5 +314,18 @@ export default class HistoryListPanel extends Mixins(BaseMixin) {
                 return value
         }
     }
+
+    createThumbnailUrl(thumbnail: FileStateFileThumbnail | undefined) {
```
meteyou, line 318:
```suggestion
    createThumbnailUrl(thumbnail: FileStateFileThumbnail) {
```

just because of the suggest above.

src/components/panels/History/HistoryListEntryJob.vue
```diff
@@ -187,40 +188,22 @@ export default class HistoryListPanel extends Mixins(BaseMixin) {
...
-        if (this.item.filename.lastIndexOf('/') !== -1) {
-            relative_url = this.item.filename.substring(0, this.item.filename.lastIndexOf('/'))
-        }
-
-        if ((thumbnail?.relative_path ?? null) === null) return false
-
-        return `${this.apiUrl}/server/files/gcodes/${encodeURI(relative_url + thumbnail?.relative_path)}?timestamp=${
-            this.item.metadata.modified
-        }`
+        return this.createThumbnailUrl(thumbnail)
```
meteyou, line 198:
```suggestion
        return thumbnail ? this.createThumbnailUrl(thumbnail) : false
```

or 

```suggestion
        if (thumbnail === undefined) return false
        
        return this.createThumbnailUrl(thumbnail)
```

i would not enter the createThumbnailUrl function, when the thumbnail was not found.

meteyou (changes requested):
i only suggested one small comment/change, otherwise the code looks good. thanks for the PR!

src/components/panels/History/HistoryListEntryJob.vue
mdziekon: Makes sense - changes applies, and in turn I could also simplify `createThumbnailUrl()` function, as now it does not have to deal with potentially `undefined` thumbnail, making nullish coalescing and optional chaining on the thumbnail object redundant.

src/components/panels/History/HistoryListEntryJob.vue
mdziekon: Changes applies, good call :+1:

src/components/panels/History/HistoryListEntryJob.vue
mdziekon: Changes applies, good call :+1:

meteyou (approved):

PR #2030, opened 2024-10-19
chore(Docker): Enable ipv6 in nginx.conf

jnimety:
I'm having connectivity issues in my dual stack environment and this seems to fixes them.

### DIFF
--- a/.docker/nginx.conf
+++ b/.docker/nginx.conf
@@ -1,6 +1,7 @@
 server {
   listen 80;
-
+  listen [::]:80;
+  
   location / {
     root /usr/share/nginx/html;
     index index.html index.htm;


meteyou:
@jnimety thx for your PR, i have to double check, if this has also some downside or it just "add ipv6 support".

pedrolamas:
@meteyou sintax is correct, that is how one enables IPv6 in nginx

https://github.com/nginxinc/docker-nginx/blob/6a4c0cb4ac7e53bbbe473df71b61a5bf9f95252f/stable/debian/10-listen-on-ipv6-by-default.sh#L63

meteyou (approved):

meteyou:
@pedrolamas thx for your feedback. i think i had a mistake here. i had the Moonraker trusted_clients in mind with the ipv6 range, but that has nothing to do with the Mainsail docker image.

PR #2057, opened 2024-11-28
fix(Webcam): make webcam view non-draggable

phm07:
## Description

This PR makes the `` tags created by the webcam preview non-draggable. They were draggable before, which would (at least in my case) lead to accidentally dragging it and uploading it to the config folder using the drag and drop function.
I was only able to test the `Mjpegstreamer` and `MjpegstreamerAdaptive`, so I didn't modify any of the other views.


meteyou:
Thx for your PR! I just added the UV4LMJPEG client, because this mode also use a img-tag.

meteyou (approved):

PR #2070, opened 2024-12-07
locale(de): update german locale

meteyou:
## Description

This PR just update the german locale.

none

none

none


src/locales/de.json
```diff
@@ -816,6 +826,9 @@
         },
         "WebcamPanel": {
             "All": "Alle",
+            "ConnectingTo": "Verbinde zu {url}",
+            "Disconnected": "Getrennt",
+            "ErrorWhileConnecting": "Fehler während dem verbinden zu {url}",
```
mryel00, line 831: Correct would be:
`Fehler während dem Verbinden zu {url}`

src/locales/de.json
```diff
@@ -1012,8 +1025,16 @@
         },
         "HeightmapTab": {
             "ColorSchemes": "Farbschemen",
+            "DefaultOrientation": "Standard Orientierung",
+            "DefaultOrientationDescription": "Standardausrichtung für die Darstellung der Höhenkarte auswählen.",
```
mryel00, line 1029: Above you call it `Standard Orientierung` and here `Standardausrichtung`. It should be the same. `Standardausrichtung` sounds better imo

mryel00 (changes requested):

src/locales/de.json
mryel00: To make it clear. It's a capitalized `V` as `dem` indicates a noun and therefore we have to capitalize it

mryel00 (approved):

PR #2075, opened 2024-12-09
fix: hide horizontal scrollbar in StartPrintDialog.vue

meteyou:
## Description

This PR hide the horizontal scrollbar in the StartPrintDialog, but the vertical scrollbar is still untouched and can be visible, when the window is too small.

none

before:
image

after:
image

none


rackrick (approved):
didn't even manged to get the horizontel scroll bar on the non fixed version ^^

PR #2129, opened 2025-01-29
fix: fix history filter reactivity

### LINKED ISSUE #2119: History is not working correctly
### Mainsail Version:

v2.13.2

### Browser:

Chrome

### Device:

Desktop PC

### Operating System:

Windows

### What happened?

On the History tab, only certain types of events related to printing are displayed. Only Klippy shutdown, Interrupted, and Klippy disconnect are shown in the list. Cancelled and Completed are not included. In the settings, Cancelled and Completed are not selected by default, but selecting them does not lead to any changes; however, the number of print jobs is displayed, meaning that the history is not lost. When the page is refreshed, the Cancelled and Completed options return to their unchecked states. A full reset with history restoration or clearing the history to start recording it anew also did not help. In early January, I updated all the packages for the first time in several months.

When selecting and checking the box, only the selected items should be displayed in the list.

Unfortunately, I have no idea.

### Additional information:

I recorded a video showing exactly what is happening: https://miruschenko.ru/disk/history-chs8bf

meteyou:
## Description

This PR fix the reactivity issue with the history list filter with status in the cogs menu of the list.

fixes #2119 

none

none


src/components/charts/HistoryFilamentUsage.vue
```diff
@@ -108,7 +110,45 @@ export default class HistoryPrinttimeAvg extends Mixins(BaseMixin, ThemeMixin) {
...
+            jobsFiltered = [
+                ...this.selectedJobs.filter(
+                    (job: ServerHistoryStateJob) =>
+                        new Date(job.start_time * 1000) >= startDate && job.filament_used > 0
+                ),
+            ]
+
+        for (let i = 0; i <= 14; i++) {
+            const tmpDate = new Date()
+            tmpDate.setTime(startDate.getTime() + 60 * 60 * 24 * i * 1000)
```
rackrick, line 134: you are adding the time here and not going back in history. this creates an empty filament usage output with future dates.

src/components/charts/HistoryPrinttimeAvg.vue
```diff
@@ -93,7 +95,34 @@ export default class HistoryPrinttimeAvg extends Mixins(BaseMixin, ThemeMixin) {
...
+            ]
+
+        if (jobsFiltered.length) {
+            jobsFiltered.forEach((current) => {
+                if (current.print_duration > 0 && current.print_duration <= 60 * 60 * 2) output[0]++
+                else if (current.print_duration > 60 * 60 * 2 && current.print_duration <= 60 * 60 * 6) output[1]++
+                else if (current.print_duration > 60 * 60 * 6 && current.print_duration <= 60 * 60 * 12) output[2]++
+                else if (current.print_duration > 60 * 60 * 12 && current.print_duration <= 60 * 60 * 24) output[3]++
+                else if (current.print_duration > 60 * 60 * 24) output[4]++
+            })
```
rackrick, line 122: well I know someone screaming magic numbers here :D
anyways I would at least add a few comments for the time ranges to make it easier undstandable in the future.

meteyou: Should be fixed with the new commit

rackrick (approved):

PR #2141, opened 2025-02-09
feat: Add search functionality to macro settings interface

mnigbur:
## Description

This PR adds search functionality to the macro settings interface

none

image

none


meteyou:
@mnigbur thx for this PR!

are you a native speaker in all these languages? normally we only add english and our native language (in my case germany). all other languages should be added by the translators.

Could you also change the layout of the search field? for example like this in the gcode file panel:
image

so it should not be the same "style/layout" as the normal "input fields for settings".

mnigbur:
> @mnigbur thx for this PR!
> 
> are you a native speaker in all these languages? normally we only add english and our native language (in my case germany). all other languages should be added by the translators.
> 
> Could you also change the layout of the search field? for example like this in the gcode file panel: image
> 
> so it should not be the same "style/layout" as the normal "input fields for settings".

@meteyou 
The translation for "search" already existed for all languages, so i prefilled it for all of them.
What's the issue with prettier? I did run the script before committing. CLRF issue?

meteyou (approved):

PR #2160, opened 2025-03-02
fix(spoolman): save spool id in lowercase variable

jakpaul:
## Description

Fixes an issue with saving spool IDs for individual tools on a multi-tool instance. Currently, no updates are written to the variables file and an error is shown when selcting a new spool for any toolhead. The cause appears to be that upper case letters (which the current implementation uses) are not allowed. See the [Klipper docs](https://www.klipper3d.org/G-Codes.html#save_variable).
This PR simply changes the variable name to lower case, resolving the problem.

Running Klipper v0.12.0-451, Moonraker v0.9.3-65

n/a

Console output displaying the error message:
grafik

n/a

Signed-off-by: Paul Gendolla <contact@gendolla.net>

### DIFF
--- a/src/components/dialogs/SpoolmanChangeSpoolDialog.vue
+++ b/src/components/dialogs/SpoolmanChangeSpoolDialog.vue
@@ -208,7 +208,7 @@ export default class SpoolmanChangeSpoolDialog extends Mixins(BaseMixin) {
         }
 
         // Set spool_id to save_variable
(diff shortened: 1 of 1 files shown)


meteyou (approved):

meteyou:
@jakpaul sry for the delay and thx for your PR!

jakpaul:
No problem. Thanks for the review

PR #2171, opened 2025-03-18
refactor: Remove duplicate PrinterStateLight definition

rcloran:
## Description

There are two identical definitions of the PrinterStateLight interface one after the other. This removes one of them.


meteyou (approved):

meteyou:
Thx for finding and fix this issue. Its just interesting, that TSC doesn't create an warning/error here.

PR #2173, opened 2025-03-22
feat: Add load cell gram scales in misc panel

rcloran:
## Description

This adds a new general purpose `MiscellaneousSensor`, with support for only load_cell gram scales. This is just a simple single line entry in the MiscellaneousPanel, with a getter to extract the value from the printer state.

With an uncalibrated load cell:

image

After calibration:

image


rcloran:
The gram scale support was added to Klipper recently, in https://github.com/Klipper3d/klipper/commit/06d65ef5ac139e2c84e7381c190ee2b1e4ec4237

@garethky , not sure if you have anything to add here, but mentioning you in case you do

meteyou (approved):

PR #2185, opened 2025-04-13
feat: Support providing moonraker with an apikey

### LINKED ISSUE #2184: Support providing moonraker with an apikey
### Requested feature:

Moonraker supports 3 differert ways to establish trust in the client:

1. The connection comes from one of the specified trusted IP addresses
2. The connection provides a secret apikey value
3. The connection logs in with a username and password.

Currently mainsail only provides affordance for the first method.  This feature request is for support for the second one.

I'm currently working on a project called "octopussy", which implements the moonraker API on top of octoprint so that you can use mainsail/fluidd/klipperscreen/... with non klipper printers.  Octoprint provides two ways of establishing trust in a client, which roughly correspond to options 2 and 3 above.  Because of this, I'm hoping to not support option 1 in octopussy, as it end-runs around octoprint's normal security measures.  That however means that mainsail can't connect to octopussy as of today, so I'm adding support for at least one of the other two methods, so that I can test mainsail against my implementation.

### Additional information:

_No response_

beelsebob:
## Description

This PR adds the authentication with api-key feature. 

Adds an optional ‘apikey’ value to the config.json file.  This is then passed through to the web socket connection, and given to moonraker to authenticate.

Fixes #2184

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -54,5 +54,6 @@ export const actions: ActionTree<RootState, RootState> = {
         if (payload.hostname) commit('socket/setData', { hostname: payload.hostname })
         if (payload.port) commit('socket/setData', { port: parseInt(payload.port.toString()) })
         if (payload.path) commit('socket/setData', { route_prefix: payload.path })
+        if (payload.apikey) commit('server/setData', { apikey: payload.apikey })
     },
 }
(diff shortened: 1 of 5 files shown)


meteyou:
I close this PR because of the discussion on the Discord. Pls create a new PR, if you have / provide a better solution.

PR #2192, opened 2025-04-16
fix(python): remove special dev changelog

Jomik:
## Description
Python package's remote hash is `update-available`, which does not make sense to show in this way.

### DIFF
--- a/src/components/panels/Machine/UpdatePanel/Entry.vue
+++ b/src/components/panels/Machine/UpdatePanel/Entry.vue
@@ -347,9 +347,6 @@ export default class UpdatePanelEntry extends Mixins(BaseMixin) {
     }
 
     get pythonChangelog() {
-        if (this.repo.channel === 'dev')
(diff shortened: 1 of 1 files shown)


meteyou:
I would just recommed to change your config instead of blocking it via config.

If you use a "default python setting" in Moonraker without setting the channel, "stable" is the default. I tested it with this example:

```
[update_manager python_test]
type: python
virtualenv: ~/klippy-env
project_name: jomik-cartographer
is_system_service: False
managed_services: klipper
info_tags: desc=Python Test
```

so I don`t think this is necessary.

Jomik:
Oh, I just saw that this was closed. Currently this results in a broken changelog for valid dev (prerelease) versions in Mainsail.
E.g. 
image
shows me https://github.com/Cartographer3D/cartographer3d-plugin/compare/275f911427f7ceadbf15097292f956fbf06b3c6f..update-available

Should it not use the changelog url?

Issue #1814, opened 2024-03-12
Add Date ISO format for display


Drizzt321:
### Requested feature:

In the Settings for the Date format selection, add ISO Date YYYY-MM-DD format option.

### Solves the following problem:

Let's me have the date in my preferred format without resorting to figuring out separating browser date format from Locale settings.

meteyou:
Already fixed with the PR #2210 but github didnt close this FR.

PR #2210, opened 2025-05-02
feat: Added more date format options

### LINKED ISSUE #2038: FR: more options for deate formats
### Requested feature:

While I am not native English speaker, I am used to english terminology in tech and therefore use English everywhere except for when it comes to date and time formats. I would like to see those in "regional format". Windows has allowed this for ages with web it's little difficult.

For some reason I am only able to pick from these date formats (MDY)
image

This as well might be "bug" because I tried setting english-US (MDY) and english-GB (DMY) as language in my browser (Firefox)
I tried executing ```alert(new Date().toLocaleString())``` in console which correctly shows time in DMY format
image
So it looks like bowser locale detection is not working properly ?

TLDR: I think it should be possible to choose whatever time format for english as many non native speakers are used to see interfaces in english while preferring to see dates in DMY, MDY or YMD, just like time can be in 12/24 

Thank you. ❤️ 

_No response_

### Additional information:

_No response_

lvsweat:
Added more date formats and added logic to handle them

## Description

This PR adds more options to choose from when selecting a timestamp/date format in general settings.

Resolves #2038, #1814 

Before:
image

After:
image
image

None


meteyou:
why do you removed 2-digits? is this option included in every locale with the other options? can you please add "dd.mm.yyyy"? this is the default german option (and i think also in other west/center europe countries)

lvsweat:
2-digits is still present, just labeled differently. Will edit the PR description shortly to show the visual changes. I have added the german options and am pushing now.

lvsweat:
@meteyou Just pinging you to make you aware of the updates for a re-review, as it has been a few weeks

lvsweat:
@meteyou No worries, and sure thing, I will likely add that option later today.

lvsweat:
@meteyou I have added the ISO date format as requested!

meteyou:
hey @lvsweat!

i saw that some formats didnt fit (mm-dd-yyyy), because it has the wrong delimiter (just check your screenshots above). i made it a little different so it will format it always the right way and we can easy add new formats.

i just found out, that some formats doesnt exists for locale format, so i think we can also remove them. what do you think of my current code and do you see all current formats as necessary / existing in the world?

Issue #1212, opened 2022-12-28
Docker user support


kjkent:
### Requested feature:

As per my admittedly limited understanding: Within Docker containers, processes run as root by default, but Docker allows for running in-container processes as a less privileged UID/GID combo. 

This has previously been accomplished by some image developers by passing the desired UID/GID via an environment variable, but docker now has the functionality to explicitly pass a `user` parameter to either `docker run` or within a Docker Compose file.

However, this `user` parameter does not currently work with the current mainsail edge image, producing logs that implicate docker-entrypoint.sh and nginx.

### Solves the following problem:

As far as I'm aware, this is good security practice and something many mainstream images allow.

### Additional information:

```
mainsail  | /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
mainsail  | /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
mainsail  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
mainsail  | 10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
mainsail  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
mainsail  | /docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
mainsail  | /docker-entrypoint.sh: Configuration complete; ready for start up
mainsail  | 2022/12/28 22:51:52 [warn] 1#1: the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
mainsail  | nginx: [warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:2
mainsail  | 2022/12/28 22:51:52 [emerg] 1#1: mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
mainsail  | nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (13: Permission denied)
```

Laikulo:
Mainsail is (for the most part) just a static web app, so that container includes nginx to act as a web server. Nginx implements it's own privilege dropping (to the `nginx` user) once it has started, as controlled by the `user` directive in it's config.

PR #2213, opened 2025-05-07
feat: add -unprivileged docker image to conform with restrired pod security standard

wilmardo:
## Description

This makes the image compliant with the restricted pod security standard:
https://kubernetes.io/docs/concepts/security/pod-security-standards/

There is major difference to the current image in that it now listens on port 8080 since ports below 1024 can only be opened by privileged users. I don't really know how to proceed with this.
Is this something that can be released with a note in the releasenotes and an update to the docs? Or I could setup a separate image which uses a `-unprivileged` postfix on the tags and keep the current image the same.
Or update the current image but setup a `-privileged` postfix with the root running image as a legacy option?

Besides that this is a good security practice, this enabled this image to run in more restricted environments where the pod security standard is enforced (Kubernetes mostly). The Talos Kubernetes enforces the restricted profile by default for example.

Fixes: #1212

## Are there any post-deployment tasks we need to perform?

A docs update is needed (which I can do) but we need to decide what the approach will be first :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,16 +14,16 @@ COPY ./ /app/
 
 RUN npm run build
 
+# set port to >1024 port for non root running
+RUN sed 's/80/8080/g' .docker/nginx.conf > .docker/nginx.conf.unprivileged
+
 #
 # Unprivileged runner stage, runs the application in nginx
 #
 FROM nginxinc/nginx-unprivileged:stable-alpine AS unprivileged
 
-COPY --link --from=builder /app/.docker/nginx.conf  /etc/nginx/conf.d/default.conf
+COPY --link --from=builder /app/.docker/nginx.conf.unprivileged  /etc/nginx/conf.d/default.conf
 COPY --link --from=builder /app/dist/ /usr/share/nginx/html/
-
-# set port to >1024 port for non root running
-RUN sed -i 's/80/8080/g' /etc/nginx/conf.d/default.conf
 
 USER nginx
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 #
 # Builder stage, builds the application in node
 #
-FROM --platform=$BUILDPLATFORM node:20-alpine as builder
+FROM --platform=$BUILDPLATFORM node:20-alpine AS builder
 
 RUN apk add zip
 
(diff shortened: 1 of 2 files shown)


wilmardo:
@pataar maybe? Since you cheched in the initial Dockerfile.

Is there anything I can do to move this forward? Otherwise I might forget and keep running my own container without upstreaming :)

wilmardo:
Updated the PR based on the feedback. Made it as backwards compatible as possible:
* squeezed the unprivileged stage between the current ones, so if someone is building it locally the end result will be the same (since the last stage will be kept which is the current rootful one)
* choose for a `sed` to set port 8080 since it is such a small change, if for some reason the differences grow it might be nice to commit a separate config file but that seems unlikely (to me at least)
* workflow updated to push the new image to a `-unprivileged` suffix. THIS IS UNTESTED. It should all work according to the docs but I had no way of testing this quickly
* set the current image push to the runner target (although this strictly isn't needed see point one) but there for clarity

Dockerfile
```diff
@@ -14,10 +14,23 @@ COPY ./ /app/
...
+#
+# Unprivileged runner stage, runs the application in nginx
+#
+FROM nginxinc/nginx-unprivileged:stable-alpine AS unprivileged
+
+COPY --link --from=builder /app/.docker/nginx.conf  /etc/nginx/conf.d/default.conf
+COPY --link --from=builder /app/dist/ /usr/share/nginx/html/
+
+# set port to >1024 port for non root running
+RUN sed -i 's/80/8080/g' /etc/nginx/conf.d/default.conf
```
pedrolamas: Maybe move this `sed` call into the builder so you don't need an extra docker layer here?

wilmardo: Yes good one, thanks! Done in 3eb25a3bc24727cfd3c254d103677acca8804762

meteyou (approved):

jpiccari:
@wilmardo Yes, it doesn't work anymore.

Appears to have broken by #2354 which attempted to fix an issue when no IPv6 addresses exist. But the script requires root permissions and breaks the unprivileged container support. At first I thought I was doing something wrong, but after realizing the functionality was recently broken, I opened #2363 to track a fix. Maybe you want to weigh in on that issue for suggestions on a fix.

PR #2212, opened 2025-05-06
refactor(Gcodefiles): refactor gcodefiles panel and table

meteyou:
## Description

This PR just refactor the gcodefiles panel & table.

## Related Tickets & Documents

none

## Mobile & Desktop Screenshots/Recordings

none

## [optional] Are there any post-deployment tasks we need to perform?

none

PR #2216, opened 2025-05-10
feat(Gcodefiles): add support for multi color gcode files 

meteyou:
## Description

This PR will add the multicolor gcode files support in the Gcode-Files page/table. This is a breakout of the PR #2089 and should be merged after #2212.

- based on the PR #2212 
- part of the PR #2089 

image

none


rackrick (dismissed):

rackrick (approved):

Dumplap:
Would you ever be interested in also including the tool corresponding to the filament? for example if you sliced a file a few months ago and used T1 and T3, or just have a file that might not use T0 you would be able to see without examining the gcode by hand.

meteyou:
Mainsail already support it, when the backend (Klipper extension) support it. Mainsail itself is just a webinterface and Mainsail Itself cannot remap Klipper commands. This have to happen in Klipper itself.

But this here is no space for discussions. so if you have a backend which this feature, pls open a feature request and descripe it with details how we could add this special setup. It have to be an open source solution!

PR #2247, opened 2025-08-17
fix(console): message formatting XSS vulnerability

jwueller:
## Description

Having HTML characters like `<` in a message will currently break it's rendering. An example would be a standard Python object `__repr__`, like it might show up in an error message:

    <klippy.extras.gcode_macro.GetStatusWrapperPython object at 0x7f67eb2cd0>

This is an XSS vulnerability. Making Klippy echo a message containing HTML could run arbitrary code in the browser of everyone currently monitoring the console.

But even if you consider this an unlikely attack vector, this will still fix quirky visual behavior in current message rendering, though.

None as far as I can tell, but I can create one for this vulnerability, if required.

### DIFF
--- a/src/plugins/helpers.ts
+++ b/src/plugins/helpers.ts
@@ -86,6 +86,10 @@ export function formatConsoleMessage(message: string): string {
     message = message.replace(/^echo:/g, '')
     // remove debug
     message = message.replace(/^debug:/g, '')
+    // escape html content
+    message = message.replace(/&/g, '&amp;')
+    message = message.replace(/</g, '&lt;')
+    message = message.replace(/>/g, '&gt;')
     // replace linebreaks with html <br>
     message = message.replace('\n// ', '<br>')
     message = message.replace(/\r\n|\r|\n/g, '<br>')


meteyou:
Closed, because i got no answer...

jwueller:
@meteyou My apologies, I must have missed this among my flurry of notifications.

I suppose this *could* be toggled, but the core issue is that we have no way to be certain about the intended message format. Either setting would produce broken output in some circumstances.

The real fix would be a native format that's attached to an individual message, so it can be rendered appropriately. However, that would likely require support from Klipper, as well as coordination with other frontends.

Since the PR no longer seems to be feasible as-is, my proposal would be to do HTML sanitization instead, to strip out any unrecognized tags (e.g. https://github.com/cure53/DOMPurify). This would allow authors to keep using common rich text elements like text styles or links, but also significantly reduce the potential attack surface.

If you're happy with the proposal, feel free to reopen this and I'll make the required changes.

jwueller:
**Addendum:** I just realized that Klipper already has a message type prefix, so maybe we should encourage standardizing around something like this?

```gcode
RESPOND TYPE=echo_html MSG="<message>"       # will likely error due to unknown value in Klipper, but:
RESPOND PREFIX="echo_html:" MSG="<message>"  # takes precedence over TYPE
RESPOND PREFIX="echo:html:" MSG="<message>"  # better backwards compatibility
```

And then consider regular `echo:` to be plain text.

Although I should add, that even in this case, HTML sanitization should still happen.

meteyou:
vouch @jwueller

meteyou:
Thx! I created a PR with DOMPurify, but this will also "delete" messages like `<klippy.extras.gcode_macro.GetStatusWrapperPython object at 0x7f67eb2cd0>`.

I also talked to Pedro from Fluidd to have the same implementation and we both dont like to change the "default behavior".

Where did you get this Klipper Python error message in the console? I thought Klipper will crash when something like that happens? Maybe we can fix this in Moonraker/Klipper to send here an escaped message?

PR #2271, opened 2025-10-11
feat(AFC): Adding togglable setting to AFC panel to display toolchange count

jimmyjon711:
## Description
- Adds togglable setting to AFC panel to display toolchange count in print stats box
  - Filament usage tooltip will still display when toolchange is displayed

None

https://github.com/user-attachments/assets/d41677a1-7012-4aa9-8a42-ae9fc5e9d840

- A fix will be implemented in AFC-Klipper-Add-On so that current toolchange value will return 0 instead of -1

### DIFF
--- a/src/components/mixins/afc.ts
+++ b/src/components/mixins/afc.ts
@@ -100,6 +100,10 @@ export default class AfcMixin extends Vue {
         return this.$store.state.gui.view.afc?.hiddenUnits ?? []
     }
 
+    get afcShowToolChangeCount(): boolean {
+        return this.$store.state.gui.view.afc?.showToolChangeCount ?? true
+    }
+
     getPrinterObject(key: string) {
         const printer = this.$store.state.printer ?? {}
         return printer[key] ?? null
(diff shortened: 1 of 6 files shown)


jimmyjon711:
> Thx for this PR! We added it a little bit different in the original AFC PR. We just show the Toolchange numbers, if the current print has just more than 10 changes. I created a small PR to add this "simpler" version here: #2295. This was faster/easier for me to add this function again, instead of changing this one. I don't like to change something in a different panel (StatusPanel), when you enable/disable something in the panel specific options (the cogs gear in the AfcPanel). So I implemented it "the old way".

@meteyou Just curious if you would be open to making the count configurable? I think I have seen some people that always want to see the count even if its less than 10.

meteyou:
@jimmyjon711 Maybe in the future, when enough users really need this feature. A configurable limit might be a better approach than a simple on/off toggle or maybe a different limit will also fix this issue.

Sometimes "less is more" — I think overthinking and overcomplicating a simple feature like this is counterproductive. Just because one or two users request something doesn't mean it should be implemented immediately. In the dev channel 
on the AFC/Boxturtle Discord, we discussed this and concluded that a limit of 10 is a good indicator to difference real multicolor prints from issues like wrong start/end gcode or simple filament changes on layers (where i also don't see any real benefit).

PR #2283, opened 2025-11-07
feat(Extruder): show PA settings for extruder_stepper

### LINKED ISSUE #845: impossible to set pressure advance values for extruder_stepper

meteyou:
## Description

This PR adds support for PA settings for extruder_stepper. It will list all steppers in the extruder panel and show which one is synced with an extruder.

- fixes #845 

grafik

none

### DIFF
--- a/src/components/panels/Extruder/ExtruderPressureAdvanceSettings.vue
+++ b/src/components/panels/Extruder/ExtruderPressureAdvanceSettings.vue
@@ -23,7 +23,11 @@
                                 dense />
                         </div>
                     </v-col>
-                    <v-col>
+                    <v-col
+                        :class="{
+                            'col-12': el.is.small || el.is.medium,
+                            'col-8': extruders.length > 1 && !el.is.small && !el.is.medium,
+                        }">
                         <pressure-advance-settings :extruder="selectedExtruder" :is-small="el.is.small" />
                     </v-col>
                 </v-row>


mcollins1290:
> Hello! If I want to disable this display feature, how would one go about doing this?

Ok I figured it out, I exported my Mainsail Settings which are stored in the Moonraker DB to a JSON file using the Settings widget, located the `'extruder'` settings and added `"showPressureAdvance":false` to the settings & values section so the complete section read `"extruder":{"showPressureAdvance":false,"showExtruderControl":true,"showExtrusionFactor":true,"showTools":true}` and then I imported my modified JSON file back in using the RESTORE button.

mcollins1290:
> @mcollins1290 this is a bug and not a feature... normally, these settings can just enable/disable it in the cogs button in the extruder panel.

Yes I was wondering that as the first place I looked was the Cogs button but I did not see a checkbox to enable/disable which I did think was odd 🤔

PR #2310, opened 2025-12-05
fix: Happy Hare - Context Menu for gates

moggieuk:
## Description
Added important functionality that was missed in merge/rework.
Specifically the addition on context sensitive menus to gates.  This is implemented with "right-click" for desktop as well as "long-click" for mobile.
It is required to be able to operation on non-selected gate. Notable example is ejecting filament from the non-current (loaded) gate.

## Related Tickets & Documents
Ticket in Happy Hare forum

## Mobile & Desktop Screenshots/Recordings
Screenshot 2025-12-05 at 10 40 52 AM

## [optional] Are there any post-deployment tasks we need to perform?
CTRL-Click the non selected gate for context specific operations menu.

Signed off by: Paul Morgan <moggieuk@hotmail.com>

PR #2312, opened 2025-12-05
fix(HappyHare): Allow unit gate wrapping

moggieuk:
## Description
This restructures the MmuUnit display to allow for wrapping when MMU has a large number of gates.  It breaks neatly retaining correctly formatted footer.

n/a

Screenshot 2025-12-05 at 1 10 20 PM

Reduce Mmu Panel width to force wrapping of gates.

Signed off by: Paul Morgan <moggieuk@hotmail.com>

### DIFF
--- a/src/components/panels/Mmu/MmuUnit.vue
+++ b/src/components/panels/Mmu/MmuUnit.vue
@@ -7,7 +7,6 @@
                 :gate-index="gateIndex - 1 + firstGateNumber"
                 :mmu-machine-unit="mmuMachineUnit"
                 :show-details="showDetails"
-                :show-context-menu="showContextMenu"
                 :unhighlight-spools="unhighlightSpools"
                 :selected-gate="selectedGate"
                 :gate-pos="gatePos(idx)"
@@ -16,15 +15,11 @@
                 v-if="hasBypass"
                 :gate-index="TOOL_GATE_BYPASS"
                 :mmu-machine-unit="mmuMachineUnit"
-                :show-context-menu="false"
                 :selected-gate="selectedGate"
                 gate-pos="R"
                 @select-gate="selectGate" />
         </div>
-        <mmu-unit-footer
-            class="pt-0 position-relative"
-            :mmu-machine-unit="mmuMachineUnit"
-            :unit-index="unitIndex" />
+        <mmu-unit-footer class="pt-0 position-relative" :mmu-machine-unit="mmuMachineUnit" :unit-index="unitIndex" />
     </div>
 </template>
 <script lang="ts">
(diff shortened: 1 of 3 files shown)


meteyou:
Thanks for the PR! Could you please rebase this onto `develop`? Currently it's based on #2310 , but since these changes are unrelated, they can be reviewed and merged independently.

src/components/panels/Mmu/MmuUnit.vue
```diff
@@ -1,26 +1,25 @@
...
                 :mmu-machine-unit="mmuMachineUnit"
                 :selected-gate="selectedGate"
+                gate-pos="R"
                 @select-gate="selectGate" />
         </div>
-        <mmu-unit-footer
-            class="pt-8 position-relative zindex-3"
-            :mmu-machine-unit="mmuMachineUnit"
-            :unit-index="unitIndex" />
+        <mmu-unit-footer class="pt-0 position-relative" :mmu-machine-unit="mmuMachineUnit" :unit-index="unitIndex" />
```
meteyou, line 23: why do you remove the z-index? I haven't test it, but i think it was necessary to have the right z order with spool and the overlap background.

src/components/panels/Mmu/MmuUnit.vue
```diff
@@ -56,6 +55,10 @@ export default class MmuUnit extends Mixins(BaseMixin, MmuMixin) {
         return this.mmuMachineUnit?.has_bypass ?? true
     }
 
+    gatePos(idx: number) {
+        return idx === 0 ? 'L' : idx === this.numGates - 1 && !this.hasBypass ? 'R' : ''
+    }
+
```
meteyou: first of all, this function is not readable. pls use guard clauses (early returns) to make it more readable.

and overall, pls rewrite this code and create a getter inside of MmuUnitGate instead of a "static prop".

src/components/panels/Mmu/MmuUnitGate.vue
```diff
@@ -52,6 +59,21 @@ export default class MmuUnitGate extends Mixins(BaseMixin, MmuMixin) {
...
+    }
+
+    gateClass(pos: string) {
+        return pos === 'L' ? 'left-gate' : pos === 'R' ? 'right-gate' : ''
+    }
+
+    gateClassContents(pos: string) {
+        const baseClass = this.gateClass(pos)
+        return baseClass ? `${baseClass}-contents` : ''
+    }
```
meteyou, line 172: same as in the MmuUnit. dont create static functions which are only run 1 time on mounting. use getters and guard clauses for more readable code.

src/components/panels/Mmu/MmuUnit.vue
moggieuk: The actual footer doesn't overlap the spools now, the area that the gate number rectangle is in does.  The footer is reduced and just spans the entire mmu unit as the final "row" without overlap.  See here:
Screenshot 2025-12-07 at 8 53 35 AM

The "fade" is accomplished in the row above so the footer color can be solid.

This appears to be the best compromise.  Without this and with the footer as before the gate numbers are "suspended" and it doesn't look good.

For reference, this is when it is not wrapped:
Screenshot 2025-12-07 at 5 11 37 PM

src/components/panels/Mmu/MmuUnit.vue
moggieuk: done.

src/components/panels/Mmu/MmuUnitGate.vue
moggieuk: done.

meteyou (approved):

PR #2316, opened 2025-12-06
fix(HappyHare): clean up of clog meter for consistent sizing

moggieuk:
## Description:
This PR cleans up the clog meter display when MMU encoder is fitted.  It ensures consistent sizing with FlowGuard meter and adds missing translation string.

n/a

Screenshot 2025-12-06 at 12 14 39 PM

Displays if MmuEncoder is fitted


src/locales/en.json
```diff
@@ -897,7 +897,7 @@
             "Ok": "Ok",
             "PreGate": "Pre-Gate",
             "RecoverState": "Fix MMU State",
-            "ShowClogDetection": "Clog Detection",
+            "ShowClogDetection": "Clog/Tangle Detection",
```
meteyou: pls do not update old translations. use a new key for a different translation. i would also not update this line in this PR, because this is important for the flowmeter PR.

src/locales/en.json
```diff
@@ -795,7 +795,7 @@
             "ButtonUnlock": "Unlock",
             "Bypass": "Bypass",
             "Cancel": "Cancel",
-            "ClogDetection": "Clog Detection",
+            "ClogDetection": "Clog/Tangle Detection",
```
meteyou: pls do not update old translations. use a new key for a different translation. i would also not update this line in this PR, because this is important for the flowmeter PR.

src/locales/en.json
meteyou: This translation have nothing todo with this PR. pls remove it or update hits branch. it should solve this issue, because this translation should be in the develop branch already.

src/locales/en.json
moggieuk: This label for for both meters.  Really the name should always have been "Clog/Tangle Detection".  I can create a new name but then the old one wouldn't be used...  please advise.

In other words this is the label for the (both) "meters"

src/locales/en.json
```diff
@@ -806,7 +806,7 @@
             "EditTtgMapTitle": "Map Tools / Filaments (Edit TTG Map)",
             "Encoder": "Encoder",
             "Extruder": "Extruder",
-            "Gate": "Gate",
+            "Flowrate": "FLOWRATE",
```
meteyou: pls make it uppersize in the frontend. you cannot be sure, that every translator also write it uppercase in the locale file.

src/locales/en.json
meteyou: make this change only in the PR, where you also need the new wording. this change is "useless" in this PR.

just think about it, when i merge this PR, but i would not merge the other PR. then the changed wording here is "wrong".

src/locales/en.json
meteyou: make this change only in the PR, where you also need the new wording. this change is "useless" in this PR.

just think about it, when i merge this PR, but i would not merge the other PR. then the changed wording here is "wrong".

src/locales/en.json
moggieuk: backed out this change.

src/locales/en.json
moggieuk: removed.

src/locales/en.json
moggieuk: done.

src/components/panels/Mmu/MmuClogMeter.vue
```diff
@@ -93,10 +95,14 @@ export default class MmuClogMeter extends Mixins(BaseMixin, MmuMixin) {
...
-    X1_END = 70 + 65 * Math.cos((60 * Math.PI) / 180)
-    Y1_END = 70 + 65 * Math.sin((60 * Math.PI) / 180)
+    X1_START = 70 + 63 * Math.cos((120 * Math.PI) / 180)
+    Y1_START = 70 + 63 * Math.sin((120 * Math.PI) / 180)
+    X1_END = 70 + 63 * Math.cos((60 * Math.PI) / 180)
+    Y1_END = 70 + 63 * Math.sin((60 * Math.PI) / 180)
+
+    get flowrateTextUpper() {
+        return this.$t('Panels.MmuPanel.Flowrate').toUpperCase()
+    }
```
meteyou: you dont need a getter to make it uppercase. just add this inline in the template section

moggieuk: done

meteyou (dismissed):

meteyou (approved):

PR #2326, opened 2025-12-07
refactor: rework the server init process to display current step

meteyou:
## Description

This PR refactor the init process of the server store. It will init all server parts sequential instead of parallel. So it can also display the current step and will show in future PRs (rework of the server components itself) the progress.

This PR is based on #2325 . So we have to merge #2325 first, then we can review this PR.

after:
grafik

none

## Release Notes

* **New Features**
  * Enhanced connection dialog displays detailed initialization progress with step-by-step feedback on system component startup.
  * Improved error messaging distinguishes between connection failures and initialization errors.

* **Improvements**
  * Better visual feedback during app startup showing which components are currently loading.
  * Expanded localization support with initialization-related messages.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


src/store/server/getters.ts
```diff
@@ -31,13 +31,21 @@ export const getters: GetterTree<ServerState, any> = {
...
+        <T = Record<string, unknown>>(section: string, fallback: T | null = null): T | null => {
+            const config = state.config ?? {}
+            if (section in config) return config[section] as T
+            return fallback
+        },
 
-        return null
-    },
+    getConfigValue:
+        (state, getters) =>
```
rackrick, line 43: state not used

meteyou: you have to follow the order with the values. so if you need getters, you have to add state, because the second value is getter... do you know a "workaround" for that?

meteyou:
@coderabbitai review

src/store/server/getters.ts
rackrick: you can mark it with an `_` as unused.
but there is not really a difference in functionality. just the ide will stop complaining.

rackrick (approved):

PR #2336, opened 2025-12-14
fix: Happy Hare eject button disabling logic

Wazzup77:
### Description

The logic for disabling the filament eject button in Happy Hare UI was copied from the preload button. This does not make sense - preload is supposed to be ran when the gate is empty and new filament is loaded, while eject if the gate is loaded and we want to remove the filament from it.
This is reflected by Happy Hare internal implementation MMU_EJECT, which exits with a "Filament not loaded" message if it does not detect filament.

### Related Tickets & Documents

None.

### DIFF
--- a/src/components/panels/Mmu/MmuControls.vue
+++ b/src/components/panels/Mmu/MmuControls.vue
@@ -71,6 +71,7 @@ import MmuMixin, {
     GATE_AVAILABLE_FROM_BUFFER,
     FILAMENT_POS_UNLOADED,
     GATE_UNKNOWN,
+    GATE_EMPTY,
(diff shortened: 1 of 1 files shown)


Wazzup77:
To add: this is further corroborated by the implementation in Fluidd:

<template #activator="{ on }">
            <v-btn
              block
              small
              :class="btnClass"
              :disabled="!klippyReady || !canSend || [GATE_EMPTY].includes(currentGateStatus)"
              :loading="hasWait($waits.onMmuEject)"
              v-on="on"
              @click="sendGcode('MMU_EJECT', $waits.onMmuEject)"
            >

meteyou (approved):

PR #2343, opened 2025-12-15
feat(HappyHare): Adds tiny numeric indicator of sensor position for Proportional Feedback sync-feedback buffers

moggieuk:
Description
This PR adds a small numeric next to buffer position for new class of proportional (analog) sensors.  The reason is the EKF (Extended Kalman Filter) for sync-feedback in Happy Hare means the buffer will almost always be centered. This indicator helps to see the minor corrections that are taking place.

Related Tickets & Documents
n/a

Mobile & Desktop Screenshots/Recordings
Screenshot 2025-12-16 at 12 50 30 AM

[optional] Are there any post-deployment tasks we need to perform?
n/a

### DIFF
--- a/src/components/panels/Mmu/MmuFilamentStatus.vue
+++ b/src/components/panels/Mmu/MmuFilamentStatus.vue
@@ -19,6 +19,9 @@
                 <rect x="3" y="0" width="30" height="40" rx="3" ry="3" fill="none" stroke-width="1.5" />
                 <path d="M-15 -4 L-6 0 L-15 4 Z" stroke-width="1" fill-opacity="0.6" />
                 <path d="M8 40 L 28 40" stroke-width="4" />
+                <text x="-22" y="4" font-size="11px" text-anchor="end" style="fill: var(--color-outline)">
+                    {{ syncFeedbackPistonText }}
+                </text>
             </g>
             <g
                 id="sync-feedback-buffer-box"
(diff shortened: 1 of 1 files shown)


src/components/panels/Mmu/MmuFilamentStatus.vue
```diff
@@ -19,6 +19,9 @@
                 <rect x="3" y="0" width="30" height="40" rx="3" ry="3" fill="none" stroke-width="1.5" />
                 <path d="M-15 -4 L-6 0 L-15 4 Z" stroke-width="1" fill-opacity="0.6" />
                 <path d="M8 40 L 28 40" stroke-width="4" />
+                <text x="-22" y="4" font-size="11px" text-anchor="end" style="fill: var(--color-outline)">
```
meteyou:
```suggestion
                <text v-if="hasFilamentProportionalSensor" x="-22" y="4" font-size="11px" text-anchor="end" style="fill: var(--color-outline)">
```

meteyou: you can use directly the if condition in the template here

src/components/panels/Mmu/MmuFilamentStatus.vue
```diff
@@ -413,6 +416,18 @@ export default class MmuFilamentStatus extends Mixins(BaseMixin, MmuMixin) {
...
+    get syncFeedbackPistonText() {
+        if (this.hasFilamentProportionalSensor) {
+            const bias = this.syncFeedbackBiasModelled
+            return bias.toFixed(2)
+        }
+        return ''
+    }
+
+    get syncFeedbackBiasModelled() {
+        return this.mmu?.sync_feedback_bias_modelled ?? 0.0
```
meteyou:
```suggestion
        return (this.mmu?.sync_feedback_bias_modelled ?? 0.0).toFixed(2)
```
and then, you can just format the plain output and dont have to use the "empty string" as fallback.

moggieuk: Thanks. Committed change.

meteyou (approved):

PR #2347, opened 2025-12-16
fix(HappyHare): Fixes animated filament position so filament doesn't go backwards

moggieuk:
## Description
This PR fixes two bugs in the animated rendering of the filament in the bowden:
1) If was possible for the filament to appear to go backwards slightly because the gate_homing_endstop was used where the extruder_homing_endstop should have been used.
2) It correctly sets the end of bowden position for ALL the "endstop" and sensor types including fixing the logic for the existence of valid toolhead sensor.  Previously this condition required sensor to be triggered which wasn't correct.

n/a

Component modified: (no visual change other than corrected position in two cases):
Screenshot 2025-12-16 at 11 00 23 PM

n/a

### DIFF
--- a/src/components/mixins/mmu.ts
+++ b/src/components/mixins/mmu.ts
@@ -320,10 +320,6 @@ export default class MmuMixin extends Mixins(BaseMixin) {
         return this.mmuSettings?.gate_homing_endstop
     }
 
-    get configExtruderHomingEndstop(): string {
-        return this.mmuSettings?.extruder_homing_endstop
-    }
-
     get canSend(): boolean {
         const idleTimeout = this.$store.state.printer.idle_timeout?.state ?? ''
 
--- a/src/components/mixins/mmu.ts
+++ b/src/components/mixins/mmu.ts
@@ -320,6 +320,10 @@ export default class MmuMixin extends Mixins(BaseMixin) {
         return this.mmuSettings?.gate_homing_endstop
     }
 
+    get configExtruderHomingEndstop(): string {
+        return this.mmuSettings?.extruder_homing_endstop
+    }
+
     get canSend(): boolean {
         const idleTimeout = this.$store.state.printer.idle_timeout?.state ?? ''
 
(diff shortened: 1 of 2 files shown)


src/components/mixins/mmu.ts
```diff
@@ -320,6 +320,10 @@ export default class MmuMixin extends Mixins(BaseMixin) {
         return this.mmuSettings?.gate_homing_endstop
     }
 
+    get configExtruderHomingEndstop(): string {
+        return this.mmuSettings?.extruder_homing_endstop
+    }
+
```
meteyou: Pls dont use the mixins, when you only use this getter one time and add a fallback, because it would not fit right know with the typing, if mmuSettings doesnt exist.

moggieuk: I'm sorry. The parallel `configGateHomingEndstop()` was in the mixin so it seemed to make sense.  I can move.

moggieuk: Moved.

src/components/panels/Mmu/MmuFilamentStatus.vue
```diff
@@ -325,6 +314,22 @@ export default class MmuFilamentStatus extends Mixins(BaseMixin, MmuMixin) {
...
+    get endOfBowdenPos() {
+        if (typeof this.toolheadSensor === 'boolean' && !this.configExtruderForceHoming) return POSITIONS.END_BOWDEN
+
+        if (
+            this.configExtruderHomingEndstop === 'none' ||
+            this.configExtruderHomingEndstop === 'collision' ||
+            this.configExtruderHomingEndstop === 'mmu_gear_touch' ||
+            this.configExtruderHomingEndstop === 'filament_compression'
+        )
+            return POSITIONS.EXTRUDER_ENTRANCE
```
meteyou:
```suggestion
        const homingEndstops = ['none', 'collision', 'mmu_gear_touch', 'filament_compression']
        if (homingEndstops.includes(this.configExtruderHomingEndstop)) return POSITIONS.EXTRUDER_ENTRANCE
```

I would write it this way. its a little bit better readable. Pls correct the name homingEndstops, if you know a better name for the array of names.

src/components/panels/Mmu/MmuFilamentStatus.vue
```diff
@@ -349,6 +354,10 @@ export default class MmuFilamentStatus extends Mixins(BaseMixin, MmuMixin) {
         return (this.$store.state.printer.configfile.config.mmu?.extruder_force_homing ?? 0) === 1
     }
 
+    get configExtruderHomingEndstop() {
+        return this.mmuSettings?.extruder_homing_endstop
```
meteyou: Pls add a fallback value here

moggieuk: Done.

meteyou (approved):

Issue #1177, opened 2022-11-30
Overlay Print information on Webcam stream


SmallStrayFox:
### Requested feature:

During a print I spend most of my time on the Webcam tab so I can keep an eye on the print and any potential issues. It would be super handy if I could also see some live print stats overlaid on the webcam feed so I can keep an eye on important things like temperatures.

### Solves the following problem:

solves having to switch between the Dashboard and Webcam tabs, when it's much more preferable to have glanceable information on the webcam feed that allows us to see any given live variable for the current print (similar to the current fps counter).

### Additional information:

ideally should be able to include any variable that's being monitored on the dash in a format that is compact but readable. graphs are nice but honestly one line of text would be more readable, e.g. you'd edit a text string like

%temp_extruder% - %temp_bed% || %temp_mcu1% - %temp_pi% || %Percent_done% - %ETA%

and it showed overlaid on the live video feed as

Extruder: 210.0°C - Heater Bed: 60.0°C || Board MCU: 38.0°C - Raspberry Pi: 32.0°C || %85 - ETA: 17:35

apphead:
Would love to see that too!
@Turbine1991 Would you mind sharing your extensions?

Hotwire:
Any movement on this? It would be SUPER handy for the time lapse to have the info too.

meteyou:
@Hotwire this have nothing Todo with timelapse.

CoffeeSquirel:
Did anything come of this?

meteyou:
@CoffeeSquirel it's still open and no pr is linked to it. So it's just an open FR.

dfbpurcell:
How do we apply the patch file?

2wenty2wo:
Also very interested in this.

PR #2389, opened 2026-01-13
feat(webcams): add printer statistic overlay for webcam streams

mspi92:
## Description

In this PR I'm implementing the possibility to overlay printer statistics into webcam streams.
Possible overlay values are:
- Extruder(s) Temperature (is/should)
- Heatbed Temperature (is/should)
- Part cooling Fan speed
- Print time
- Estimate time remaining
- Flow rate (in mm^3/s
- Movement speed (in mm/s)
- Layer count (n of total)
- ETA

The information is conditionally hidden if no value is available (i.e. no print is running). 

You can choose between the position where the items are shown (top-left, top-right, bottom-left, bottom-right).

Additionally the overlays can be tinted with a background color. By setting alpha component to 0, the background-color is transparent. This can be useful for certain setups to avoid black printer parts obfuscating the black fontsize.

The settings page webcam preview shows dummy values, so you don't need to have a print-running to see the position of all overlayed values.

See screenshots for concrete details.

Unrelated note: `npm run format` reformatted the additionalSensor array in variables.ts.

Adds: https://github.com/mainsail-crew/mainsail/issues/1177

Settings Page Webcam Preview
Settings page webcam preview
Settings
Settings Page Settings
Dashboard Webcam View
Bildschirmfoto 2026-01-13 um 13 08 00


meteyou:
thx for your PR! I just have some small ideas to improve this function.

1. Can you add a checkbox to enable/disable all overlays at once and hide all options below? Like the "nozzle crosshair" function. It also hids/show the suboptions with this checkbox

2. do you think it make sense to combine some options? for example just have 1 checkbox for "Temperatures" instead of "extruder, bed, ..."?

3. maybe an icon would which change to a different one for the position (left-top, right-top, ...). For example an arrow, which rotate on each click on it, instead of the select box. less keys to translate + it needs less space and the label has more space.

4. would it make sense to add an option to only show this data only in the webcam page, not on the dashboard?

mspi92:
Hey, thanks for your first review :)
1. Will add those options.
2. Grouping categories removes the possibility to choose the position of extruder temp vs. bed independently though. Generally the menu itself has quite a lot of options right now, but generally it is a one time operation configuring it and than it is better to have the option to choose something. I also tested this PR during a Print and discovered myself looking at those values specifically.
3. Agree, I18n is always a topic and sign language is kind of universal :D Luckily Vuetify covers us here I guess :D 
4. I must admit that I implemented this feature in the components to hide values for exactly that reason, but didn't come around executing this. Fix will come as well here

mspi92:
From the discussion above:
1. Enable/disable with Hiding
2. I didn't group the settings because of the aforementioned points.
3. Drowdowns are replaced by icons, makes it indeed more optically pleasing
4. Dashboard Webcam Widgets hides the extra Info, only the main Webcam Page shows it now :)

Additionally i included Icons next to the texts

See the screenshot below for reference
Settings Page redesign
Settings Page

meteyou:
ohh! this looks a lot of nicer!

is it possible to make the "arrows" (orientation buttons/icons) a little bit bigger? i would just use the same estimate config like you can set for the dashboard. then you dont need "submenu" here. i like the liddle icons before the label text!

mspi92:
I fixed the styling of the position buttons (see screenshot).
Fixed Settings

For the estimation config again we have 2 values, also shown in the dashboard next to each other. I tend to use the Slicer Estimate, as it is more accurate in the first 50% of the print and just want to give any user the flexibility.
In my opinion a 3D Printer and a Software like mainsail is an expert-tool and a few options more don't harm.

Let me know if I should change it though, your call ;)

mryel00:
> Dashboard Webcam Widgets hides the extra Info, only the main Webcam Page shows it now :)

1. About that, I just tested the PR and the current implementation shows a different view on the cam settings page than you will end up seeing on the webcam page. I'm not entirely sure, if scaling this to size (really small text in webcam settings) make any sense, but currently it's quite a mess in the webcam setting, if you check a lot of options, but it will be ok on the webcam page, as you can see on the screenshot.

2. Some note on the order of the overlay options, you should order them by "able to see out of print" and "only able to see during print". That way it will be easier for the user to understand when you will see which one and the new "during print" attributes will be shown at the same location (below or on top of) in respect to the "out of print" attributes. I hope this makes sense.

3. A small QOL idea, add a right click option that rotates counter clockwise.

Besides those points, I think this is a nice feature. Thank you for contributing :heart: 

grafik

mspi92:
I finally came around to implementing all fixes reviewed by CodeRabbit.
Additionally I implemented a few of your suggested fixes @mryel00:

> 
> 1. About that, I just tested the PR and the current implementation shows a different view on the cam settings page than you will end up seeing on the webcam page. I'm not entirely sure, if scaling this to size (really small text in webcam settings) make any sense, but currently it's quite a mess in the webcam setting, if you check a lot of options, but it will be ok on the webcam page, as you can see on the screenshot.
> 
> 2. Some note on the order of the overlay options, you should order them by "able to see out of print" and "only able to see during print". That way it will be easier for the user to understand when you will see which one and the new "during print" attributes will be shown at the same location (below or on top of) in respect to the "out of print" attributes. I hope this makes sense.
> 
> 3. A small QOL idea, add a right click option that rotates counter clockwise.

To 1: 
Bildschirmfoto 2026-03-13 um 19 03 20
The Font-Size gets readjusted to better reflect the "real size and how it looks like. This mitigates the text overflow.

To 2:
Bildschirmfoto 2026-03-13 um 19 02 50
The settings are grouped and captioned according to what will be happening in the "real view"

To 3:
To my knowledge it isn't easily possible to capture a right click event in a Vue Button as it defaults to the Browser Context Menu. For this small QoL Item I don't want to go the full route of hooking into the browsers click handler to capture those click events.

Thanks for reviewing and I'm happy for comments :)

PR #2394, opened 2026-01-14
fix: ETA day calculation

mspi92:
## Description

Sorry to all americans out there, I'll use the 24h time format for my explanation :D 
When starting a print at 22:00 with a time estimate of 3h the resulting shown ETA was 01:00.
From Bambu Studio I got used to the ETA showing 1:00 +1 indicating its happening on the next day.
Some digging revealed that code parts like this are integrated, but triggering only for runtimes over 24h which is unintuitive to me.
In case the behavior is as expected, feel free to close this PR without notice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved printer ETA display formatting for better readability, now showing time in user-preferred format with day offset indicators (e.g., "+1" for tomorrow) when applicable.

* **Tests**
  * Added comprehensive test coverage for ETA formatting scenarios including edge cases.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.github/scripts/compare-bundle-sizes.sh
+++ b/.github/scripts/compare-bundle-sizes.sh
@@ -96,8 +96,8 @@ small_css_count=0
 threshold=51200
 
 while read -r normalized; do
-  base_size=$(grep "^$normalized " "$base_sizes" | awk '{print $2}' | head -1)
-  pr_size=$(grep "^$normalized " "$pr_sizes" | awk '{print $2}' | head -1)
+  base_size=$(grep "^$normalized " "$base_sizes" 2>/dev/null | awk '{print $2}' | head -1 || true)
+  pr_size=$(grep "^$normalized " "$pr_sizes" 2>/dev/null | awk '{print $2}' | head -1 || true)
 
   base_size=${base_size:-0}
   pr_size=${pr_size:-0}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
           name: mainsail-latest.zip
           path: dist/mainsail.zip
 
-  test:
+  cypress:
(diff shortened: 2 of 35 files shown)


meteyou:
Hey! We tried to rebuild the same behavior like on flying tickets. Maybe we make a mistake here. I will try do dig in the internet, if I can find something "official" for fly times/ETA.

meteyou:
Refactored the ETA formatting logic to fix several issues (some of them, was already before in my old code):

**Bug fixes:**
- **Infinite loop:** The while-loop calculating day difference could run indefinitely if ETA was in the past or far in the future. Replaced with direct date arithmetic.
- **12-hour format AM/PM:** Fixed incorrect AM/PM assignment. Previously, 12:xx was always shown as PM regardless of noon/midnight.
- **Past ETA handling:** Now returns `'--'` when ETA is in the past instead of displaying invalid times.

**Code improvements:**
- Replaced magic number `24 * 60 * 60 * 1000` with `MS_PER_DAY` constant
- Simplified day difference calculation by mutating local Date objects directly

I havn't tested this code until now. So please double check it, then we can merge it.

mspi92:
Sounds like a wonderful case to integrate vitest and write some unittests :D This also makes it much easier tampering with the system time :D 

This still needs to be integrated into Github Actions though

meteyou:
ohh! thats awesome!

would it be possible to add vitest in an own PR with a github action to check on each PR and add in this PR only the specific tests?

mspi92:
Opened a seperate PR with: https://github.com/mainsail-crew/mainsail/pull/2396 to add Vitest Environment. Once the environment PR is merged I will adjust this PR accordingly

meteyou (approved):

PR #2427, opened 2026-02-08
fix(HappyHare): Fixes sync-feedback buffer state output in visualization and supports analog sensors

moggieuk:
## Description
This PR fixes the incorrect display of sync-feedback buffer state on certain sensor types. Happy Hare supports dual-switch sensors, compression-only switch, tension-only switch as well as newer analog (proportional) sensors.  Previously the code looked at sensor switch state for these annotations. Now it uses the Happy Hare support status which allows for configuration on the zone for tension/compressed states.  It also now works correctly with analog/proportional sensors.

n/a

The buffer rendering is unchanged except for the correct display of "Neutral", "Compressed", or "Tension" status under all conditions:
sync_feedback_fixes_after

n/a


meteyou (approved):

mswarbrick89:
this issue isnt fixed, my sensor still shows normal text even under tension or compression.

PR #2434, opened 2026-02-09
feat(tempgroups): add user-organizable temperature tabs

### LINKED ISSUE #2376: User organizable temperature tabs
### Requested feature:

Hi,

I am writing this FR by request of meteyou after a discussion on Discord. 

The idea is to have multiple tabs in the temperatures window, each with their own temperature graph. For example, in my case I would do:

1. Print temperatures: extruder, heater, chamber temp average
2. Chamber temps: this would include all of the thermistors separately. Having them on their own graph makes it easier to read, since the extruder temp is not included, and so the graph can be scaled differently.
3. Electronics. This is information I would still like to have access to occasionally, but do not need to see all the time.

I imagine the organization of the tabs could be done in a similar way to how groups can be created for macros.

When you have a lot of temperature sensors, it becomes hard to see important information at a glance. The many thermistors form a long list, pushing the graph out of view on some screens. Some sensors, such as e.g. the Cartographer coil can't be renamed. And temperature controlled fans are always at the top of the list. Hiding certain temperatures, although possible in some cases, is not a great solution.
Image

### Additional information:

_No response_

Ragilo:
## Summary

Adds ability to organize temperature sensors into custom groups (tabs), addressing the issue of cluttered temperature panels with many sensors.

- Create, edit, and delete temperature groups via Settings → Temp Groups
- Drag & drop to reorder groups and sensors within groups
- Optional "All" tab showing all sensors (default behavior preserved)
- Custom display names for sensors (solves issue with non-renamable sensors like Cartographer)
- Per-group chart visibility toggle
- Graph Y-axis scales appropriately for filtered sensor range
- Full backwards compatibility - no tabs shown without configured groups

### Screenshots
image
image
image
image

## Test plan

- [x] Create a new temperature group
- [x] Add sensors to the group
- [x] Verify only grouped sensors appear in the tab
- [x] Verify graph filters correctly for the active tab
- [x] Test drag & drop reordering of groups and sensors
- [x] Test custom display names for sensors
- [x] Verify "Show All tab" toggle works
- [x] Verify settings persist after browser reload
- [x] Test on mobile (responsive tabs)

Closes #2376

### DIFF
--- a/src/components/TheSettingsMenu.vue
+++ b/src/components/TheSettingsMenu.vue
@@ -81,6 +81,7 @@ import SettingsGCodeViewerTab from '@/components/settings/SettingsGCodeViewerTab
 import SettingsEditorTab from '@/components/settings/SettingsEditorTab.vue'
 import SettingsTimelapseTab from '@/components/settings/SettingsTimelapseTab.vue'
 import SettingsNavigationTab from '@/components/settings/SettingsNavigationTab.vue'
+import SettingsTempgroupsTab from '@/components/settings/SettingsTempgroupsTab.vue'
 
 import Panel from '@/components/ui/Panel.vue'
 import {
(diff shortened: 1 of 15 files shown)


CarnageMarkus:
Can this be merged please ?

printed-hands:
Please merge this. It useful for those with at least 12 temperature sensors in their machine.

PR #2471, opened 2026-04-12
fix(socket): resync after long tab inactivity to avoid replay catch-up

LucianBuzzo:
*🐦‍⬛ Co-authored with Gravious.*

## Summary
- Adds a long-inactivity resume path that performs a full state resync.
- Suppresses incremental websocket notifications during resync to avoid catch-up replay artifacts.

## Problem
When the Mainsail UI is backgrounded/throttled (e.g. embedded webview or inactive tab) for long periods, returning can produce a burst of incremental updates that appear as rapid replay/fast-forward.

## Approach
- Track tab hidden/visible transitions.
- If hidden gap exceeds 180s, enter `socket.isResyncing` mode.
- While resyncing, ignore noisy incremental notify events.
- Trigger fresh state initialization via `printer/init` and `server/init`.
- Exit resync mode once complete.

## Notes
- Short tab switches are unaffected.
- This intentionally prefers authoritative snapshot state over stale incremental replay after long inactivity.

## Testing
- ESLint run on changed files passed.
- Manual code-path verification for visibility gap handling.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -64,6 +64,8 @@ import { AppRoute } from '@/routes'
     },
 })
 export default class App extends Mixins(BaseMixin, ThemeMixin) {
+    private hiddenAt: number | null = null
+    private readonly longVisibilityGapMs = 180_000
     get title(): string {
         let title = this.$store.getters['getTitle']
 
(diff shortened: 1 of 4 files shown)


meteyou:
We are right now refactor the complete store. So maybe this fix will be dropped after the refactoring... I will review it after the refactoring it.

PR #2474, opened 2026-04-18
fix(JobQueue): batch adding file from subdir to queue

Theta-Dev:
## Description

I noticed that I could not use the option "Add batch to queue" on the gcode files tab if the file was located in a subdirectory.
I looked into the requests and found that Mainsail sends the filename preceeding with a slash (example: "/FilamentSwatch/favorite-filament-prusa-petg-white_0.4n_0.2mm_PETG_MK4S_20m.gcode". This causes Moonraker to return a "File not found" error.

I fixed it in a similar way to the StartPrintDialogThumbnail, i.e. added a getter to get the path without leading slash.


meteyou:
Thx for your PR! I will review it after my vuex store refactoring. I will not create merge conflicts in this time.

meteyou (approved):

PR #2509, opened 2026-05-28
fix(zoffset): guard isEndstopProbe against missing endstop_pin

### LINKED ISSUE #2508: ZoffsetMixin.isEndstopProbe throws TypeError when stepper has no endstop_pin
**Describe the bug**

`ZoffsetMixin` (`src/components/mixins/zoffset.ts`) crashes when the active stepper's config has no `endstop_pin`. The `endstop_pin` getter intentionally returns `null` in that case, but `isEndstopProbe` then calls `.replaceAll(' ', '')` on it:

```js
get endstop_pin() {
    return this.settings[this.stepper_name]?.endstop_pin?.trim() ?? null
}
get isEndstopProbe() {
    return this.endstop_pin.replaceAll(' ', '').search('probe:z_virtual_endstop') !== -1
}
```

`null.replaceAll` throws `TypeError: Cannot read properties of null (reading 'replaceAll')`. Because `showSaveButton` evaluates `isEndstopProbe` only after `z_gcode_offset !== 0`, the error is latent until a non-zero Z offset exists, at which point the throw causes Vue to drop the entire Z-offset control from the Toolhead panel.

**To reproduce**

Run a Klipper-compatible backend whose `configfile.settings.stepper_z` does not include `endstop_pin`, set a non-zero gcode Z offset (e.g. after probe-based leveling), and open the dashboard. The Z-offset control disappears and the console shows the TypeError.

**Expected behavior**

A missing `endstop_pin` should be treated as "not an endstop probe" rather than crashing the component.

**Context**

Seen with a Klipper-API-compatible firmware (Anycubic GoKlipper) that omits `endstop_pin` from `stepper_z`. Vanilla Klipper always populates it, which is why this hasn't surfaced before. A null guard in `isEndstopProbe` resolves it.

martinbogo:
## Description

`ZoffsetMixin.isEndstopProbe` calls `.replaceAll(' ', '')` on `endstop_pin`, but the `endstop_pin` getter returns `null` when the active stepper config has no `endstop_pin`:

```js
get endstop_pin() {
    return this.settings[this.stepper_name]?.endstop_pin?.trim() ?? null
}
get isEndstopProbe() {
    return this.endstop_pin.replaceAll(' ', '').search('probe:z_virtual_endstop') !== -1
}
```

`null.replaceAll` throws `TypeError: Cannot read properties of null (reading 'replaceAll')`. Since `showSaveButton` evaluates `isEndstopProbe` only after `z_gcode_offset !== 0`, the error is latent until a non-zero Z offset exists, at which point the throw causes Vue to drop the entire Z-offset control from the Toolhead panel.

This coalesces `endstop_pin` to an empty string in `isEndstopProbe`, so a missing pin is treated as "not a probe" rather than crashing the component.

Fixes #2508

- #2508

## Steps to test / reproduce

Run a Klipper-API-compatible backend whose `configfile.settings.stepper_z` has no `endstop_pin`, set a non-zero gcode Z offset (e.g. after probe-based leveling), and open the dashboard. Before this change the Z-offset control disappears and the console shows the TypeError; after it, the control renders normally.

Seen in the wild with Anycubic GoKlipper, which omits `endstop_pin` from `stepper_z`. Vanilla Klipper always populates it, which is why this hasn't surfaced before.

### DIFF
--- a/src/components/mixins/zoffset.ts
+++ b/src/components/mixins/zoffset.ts
@@ -38,7 +38,9 @@ export default class ZoffsetMixin extends Vue {
     }
     get isEndstopProbe() {
         // remove spaces and search for probe:z_virtual_endstop
-        return this.endstop_pin.replaceAll(' ', '').search('probe:z_virtual_endstop') !== -1
+        // guard against a missing endstop_pin (null) which some Klipper-compatible
+        // backends omit from stepper config; null.replaceAll would otherwise throw
+        return (this.endstop_pin ?? '').replaceAll(' ', '').search('probe:z_virtual_endstop') !== -1
     }
 
     get existZOffsetApplyProbe() {


meteyou:
Thanks for your PR, but this issue was also fixed with PR https://github.com/mainsail-crew/mainsail/pull/2494 which i merged first.

PR #2511, opened 2026-05-30
fix(afc): pull remaining weight from Spoolman when enabled

paxx12:
## Description

This PR fixes the AFC lane weight display to pull `remaining_weight`
from Spoolman when Spoolman is enabled, instead of always using the
`weight` value from the AFC lane state.

When Spoolman is configured, `spoolRemainingWeight` now reads
`spool.remaining_weight` from the Spoolman spool data. This keeps
the displayed weight and spool fill percentage in sync with
Spoolman's tracking (which accounts for filament consumed across all
prints), rather than relying solely on the value AFC reports for the
lane.

Falls back to `lane.weight` when Spoolman is not enabled or no spool
is linked to the lane.

- https://github.com/fluidd-core/fluidd/pull/1855

No visual layout change — only the weight value source changes when
Spoolman is active.

Signed-off-by: paxx12 <paxx12dev@gmail.com>

### DIFF
--- a/src/components/panels/Afc/AfcPanelUnitLaneBody.vue
+++ b/src/components/panels/Afc/AfcPanelUnitLaneBody.vue
@@ -101,9 +101,10 @@ export default class AfcPanelUnitLaneBody extends Mixins(BaseMixin, AfcMixin) {
     }
 
     get spoolRemainingWeight() {
-        if (this.afcExistsSpoolman && this.spool?.remaining_weight != null) {
+        if (this.afcExistsSpoolman && this.spool) {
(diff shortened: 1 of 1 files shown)


paxx12:
@jimmyjon711 Following our conversation about this change. Please review and approve.

jimmyjon711 (approved):
Tested with spoolman and change weight manually that AFC tracks, acts like it should and only pulls in spoolmans weight. Switch to not having spoolman and weight tracks AFC weight correctly.

meteyou (approved):

PR #2512, opened 2026-05-30
feat(afc): support multiple tools per lane via `map` array

paxx12:
## Description

Broadens `AfcLaneState.map` from `string | null` to `string | string[] | null` so the backend can assign multiple tools to a single lane. This is backwards-compatible — existing firmware that sends a plain string continues to work without changes.

This is to provide support for mapping logical tools (provided by slicer) to physical topology (actual filament loaded): 

- https://github.com/AFCProject/AFC-Klipper-Add-On/issues/605
- https://github.com/fluidd-core/fluidd/pull/1856

The behavior of mapping, whether the tool is remapped, or moved depends purely on how `AFC` executes the `SET_MAP`. This basically allows to visualise `map: string[]` values, and select from those values.

Screen Recording 2026-05-30 at 17 35 26

### DIFF
--- a/src/components/dialogs/AfcUnitLaneMappingToolDialog.vue
+++ b/src/components/dialogs/AfcUnitLaneMappingToolDialog.vue
@@ -58,7 +58,8 @@ export default class AfcUnitLaneMappingToolDialog extends Mixins(BaseMixin, AfcM
 
     get mappedTools(): string[] {
         const map = this.lane.map
-        if (map == null) return []
+        if (!map) return []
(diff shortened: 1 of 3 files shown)


jimmyjon711 (approved):
Looks good to me, tested with older AFC that does not have "Virtual Tool" mapping yet.

https://github.com/user-attachments/assets/405721bc-e711-4328-abfd-faf567971ece

meteyou (approved):

PR #2515, opened 2026-06-04
feat: decouple websocket protocol from web UI origin

jothan:
## Description

This PR decouples Mainsail's WebSocket protocol (`ws://` vs `wss://`) from the browser's `window.location.protocol`. 

Previously, Mainsail forced the connection protocol to match the web interface's hosting environment. This prevented users running Mainsail locally over plain HTTP from successfully connecting to remote printers that enforce secure (`wss`) connections.

**Key Changes:**
* **Explicit Protocol Construction:** Rewrote the `getUrl` and `getWebsocketUrl` getters in `src/store/socket/getters.ts`. `getUrl` no longer generates protocol-relative URLs (`//hostname...`), which previously caused API fetch requests to blindly inherit the UI's HTTP protocol. It now explicitly respects the printer's configured protocol flag.
* **Remote Printer UI:** Added a "Secure Connection" checkbox to the "Add/Edit Printer" dialogs (`TheSelectPrinterDialog.vue` and `SettingsRemotePrintersTab.vue`). 
* **Mixed Content Fallbacks:** The secure connection flag defaults to off when Mainsail is hosted on HTTP. If the user loads Mainsail over HTTPS, the checkbox is intentionally hidden and `wss` is forced, protecting the app from violating browser Mixed Content Policies.
* **State Propagation:** Ensures the custom `protocol` flag is appropriately persisted to local storage and hydrated correctly when editing existing printers.

*(Disclaimer: This feature was "vibe coded" with the assistance of an AI agent, but has been tested manually against a remote Moonraker instance requiring auth/WSS. Please scrutinize during code review!)*

N/A

Screenshot from 2026-06-04 19-54-11

None.

### DIFF
--- a/src/components/TheSelectPrinterDialog.vue
+++ b/src/components/TheSelectPrinterDialog.vue
@@ -72,13 +72,24 @@
                             </v-col>
                             <v-col class="col-4">
                                 <v-text-field
-                                    v-model="dialogAddPrinter.port"
+                                    v-model.number="dialogAddPrinter.port"
                                     :rules="[(v) => !!v || $t('SelectPrinterDialog.PortRequired')]"
                                     :label="$t('SelectPrinterDialog.Port')"
                                     hide-details="auto"
                                     required
                                     outlined
                                     dense />
+                            </v-col>
+                        </v-row>
+                        <v-row v-if="!isHttps" class="mt-0 pt-0">
+                            <v-col class="py-0">
+                                <v-checkbox
+                                    v-model="dialogAddPrinter.secure"
+                                    :label="$t('SelectPrinterDialog.SecureConnection')"
+                                    hide-details="auto"
+                                    class="mt-1"
+                                    dense
+                                    @change="onSecureChange('add')" />
                             </v-col>
                         </v-row>
                         <v-row v-if="showOptionalSettings">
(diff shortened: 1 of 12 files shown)


meteyou:
Thanks for the work here. Before going further, I'd like to understand the use case better: when would you actually serve Mainsail over plain HTTP but connect to Moonraker over `wss://`?

In a real deployment, if you've already put SSL/TLS in front of Moonraker, Mainsail is typically served over the same reverse proxy and therefore over HTTPS as well, so both ends are encrypted and this decoupling isn't needed. The only case I can see where you'd have HTTP-served Mainsail + `wss://` Moonraker is the local dev environment (`npm run dev`), and I don't think that alone justifies adding this to the codebase.

If you have a concrete production scenario in mind, I'm happy to hear it.

On top of that, the same situation as in #2516 applies: I'm currently doing a complete refactor of the stores and the init process (`feat/rework-init-process`), and native auth is being designed as part of that rework — which touches exactly the socket/URL and protocol-handling code this PR changes. So even setting the use case aside, a PR of this size would conflict heavily and break once the refactor lands.

jothan:
> Thanks for the work here. Before going further, I'd like to understand the use case better: when would you actually serve Mainsail over plain HTTP but connect to Moonraker over `wss://`?
> 
> In a real deployment, if you've already put SSL/TLS in front of Moonraker, Mainsail is typically served over the same reverse proxy and therefore over HTTPS as well, so both ends are encrypted and this decoupling isn't needed. The only case I can see where you'd have HTTP-served Mainsail + `wss://` Moonraker is the local dev environment (`npm run dev`), and I don't think that alone justifies adding this to the codebase.
> 
> If you have a concrete production scenario in mind, I'm happy to hear it.
> 
> On top of that, the same situation as in #2516 applies: I'm currently doing a complete refactor of the stores and the init process (`feat/rework-init-process`), and native auth is being designed as part of that rework — which touches exactly the socket/URL and protocol-handling code this PR changes. So even setting the use case aside, a PR of this size would conflict heavily and break once the refactor lands.

@meteyou I indeed implemented it to make testing my use case easier and to be able to test a secure and plain connection at the same time. I think the slight complexity added might be worth it if browser-based integration tests are added someday.

PR #2520, opened 2026-06-10
feat: service worker navigation scoping and auth proxy redirect detection

jothan:
## Description

This PR fixes two issues that arise when Mainsail is deployed behind an authentication proxy (e.g. Authelia) alongside other apps on the same subdomain.

### Service worker navigation scope (`vite.config.ts`)

The Workbox-generated service worker previously intercepted **all** navigation requests and served Mainsail's `index.html` for every path not in the denylist. This broke co-hosted apps (e.g. Spoolman at `/spoolman`, Authelia at `/authelia`).

- Added a `navigateFallbackAllowlist` so the SW only handles paths that Mainsail actually owns (`/`, `/allPrinters`, `/cam`, `/console`, `/heightmap`, `/files`, `/viewer`, `/history`, `/timelapse`, `/config`, `/settings`). Any other path now falls through to the real server.
- The existing denylist for Moonraker backend paths is retained as a belt-and-suspenders measure.
- Added a `NetworkFirst` runtime caching rule for navigation requests to prevent auth redirects (301/302) from being cached by the service worker.

### Auth proxy redirect detection on WebSocket failure (`webSocketClient.ts`)

When an auth proxy intercepts a WebSocket upgrade request, it returns an HTTP redirect. The browser's WebSocket API cannot follow redirects — it sees a non-101 response and fires `onerror`/`onclose` with `wasClean=false` (code 1006), with no HTTP status or redirect URL exposed to JavaScript. Mainsail previously treated this identically to a temporary network failure and would loop through all reconnect attempts before showing a generic "Cannot connect" dialog.

**Probe mechanism:** On the first unclean WebSocket close, an HTTP probe is launched in parallel with the reconnect loop:

- `fetch(probeUrl, { redirect: 'follow' })` where `probeUrl` is the WebSocket URL rewritten to `https://.../server/info`
- If `response.redirected === true`, the user is navigated to the auth portal with the return URL rewritten to point back to Mainsail
- If the fetch throws or returns without a redirect, the reconnect loop continues normally

**Cancellation:** `probeController` (`AbortController | null`) governs the probe lifecycle — aborted on fresh `connect()` calls and successful opens to prevent stale probes from navigating the user away after recovery.

**Bug fix:** Also fixes a pre-existing bug where `reconnects` was never reset when `connect()` was called manually ("Try Again"), so a second manual attempt after exhausting all retries would immediately give up without retrying.

_No related issues yet._

_No visual changes — this PR only affects service worker configuration and WebSocket connection logic._

Users behind auth proxies should clear their browser's service worker cache after updating to ensure the old SW (which cached all navigation routes) is replaced. This can be done via the browser DevTools (Application → Service Workers → Unregister) or will happen automatically on the next SW update cycle.

### DIFF
--- a/src/plugins/webSocketClient.ts
+++ b/src/plugins/webSocketClient.ts
@@ -16,6 +16,14 @@ export class WebSocketClient {
     store: Store<RootState> | null = null
     waits: Wait[] = []
     heartbeatTimer: number | null = null
+    /** Pending reconnect timer handle; cleared on each fresh connect() call. */
+    private reconnectTimer: number | null = null
+    /**
+     * AbortController for the in-flight probe fetch; also doubles as the
+     * "probe already launched" flag (null = not yet probed this sequence).
+     * Aborted and nulled on fresh connect(); nulled on successful open.
+     */
+    private probeController: AbortController | null = null
 
     constructor(options: WebSocketPluginOptions) {
         this.url = options.url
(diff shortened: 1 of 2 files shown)


meteyou:
Sry but the complete PR is just vibe coded. More comments than code and multiple scopes in 1 PR.

Pls try it again with your brain and clean code. I cannot merge something like that.

PR #2526, opened 2026-06-12
Master

frarthur:


mryel00:
@frarthur Fix your AI agent. It's producing unnecessary noise

PR #2527, opened 2026-06-12
feat(maintenance): add datetime picker for start_time

frarthur:
## Description

Adds a datetime picker to the Add and Edit maintenance dialogs, allowing users to set a custom start_time instead of always using the current time.

## Changes

- **HistoryListPanelAddMaintenance.vue**: Add datetime-local input, selectedDateTime property, validated date parsing with Date.parse + Math.floor
- **HistoryListPanelEditMaintenance.vue**: Same datetime picker, pre-populated from existing entry's start_time

### DIFF
--- a/src/components/dialogs/HistoryListPanelAddMaintenance.vue
+++ b/src/components/dialogs/HistoryListPanelAddMaintenance.vue
@@ -29,6 +29,17 @@
                 </v-row>
                 <v-row>
                     <v-col>
+                        <v-text-field
+                            v-model="selectedDateTime"
+                            type="datetime-local"
+                            :label="$t('History.StartTime')"
+                            outlined
+                            dense
+                            hide-details="auto" />
+                    </v-col>
+                </v-row>
+                <v-row>
+                    <v-col>
                         <settings-row :title="$t('History.Reminder')">
                             <v-select
                                 v-model="reminder"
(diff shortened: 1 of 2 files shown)


meteyou:
Just AI-Slop and no reaction on the bot review... I also don't think the maintenance entry should have a edit able start_time.

frarthur:
For the timezone, Using datetime-local + Date.parse() in the browser's local timezone is intentional.
The stored epoch correctly represents the absolute moment the user selected, this is the right behavior for a local/desktop UI.
The use case here is CMMS-style: recording past maintenance events (nozzle changes, part replacements, etc.) with their actual wall-clock time. Since the app is local to the printer, cross-timezone multi-user concerns don't apply.

I Fixed toDatetimeLocalString now formats HH:mm:ss and the input has step="1".

And toDatetimeLocalString is now in BaseMixin instead of being duplicated in both dialogs

meteyou:
the time will be used from the client browser and not from the SBC. Mainsail is a webapp and is only executing in the browser from the user. there is no "mainsail server" on printer itself. so "new Date" is the timezone from the user, not from the printer/sbc.

But your description is complete different from the PR itself. The PR adds a custom input field to enter a different date/time to a entry...

Is it possible, that this PR is just AI slop and you doesnt know what your clanker add here?

frarthur:
You're right, I should have opened an issue first. My bad. Happy to open an issue and discuss the use case properly if there's interest later.

meteyou:
Yes pls! That would be nice.

Pls open an issue with a detailed description of your issue right now.

frarthur:
Sorry, I should have read the contributing guidelines before opening the PR. I'm not super experienced with GitHub workflows yet, lesson learned.

I'll open an issue. Thanks for the patience.

PR #2611, opened 2026-07-26
feat(dashboard): rearrange panels via drag and drop

Hannott:
## Description

This PR adds drag and drop rearranging of panels directly on the dashboard, as an alternative to the sortable lists in the settings dialog.

- Panels are grabbed by the icon in their header and can be moved within their column or into another column of the current viewport. Grabbing the icon starts the drag immediately; the rest of the header (buttons, collapse) behaves as before.
- While dragging, the panel itself stays rendered as a live, slightly dimmed preview at the current drop position, and the element following the cursor is a translucent box with just the panel header, which follows the size of the preview. The columns are highlighted as drop targets, and hover on elements below the cursor is suppressed for the duration of the drag.
- The new order is written to the same `dashboard.<viewport>Layout<n>` settings used by the settings dialog, so both stay in sync. Because the dashboard only renders visible and currently available panels, the reordered list is merged back into the stored layout, so hidden panels and panels of currently unavailable services (e.g. Spoolman offline) keep their position relative to their stored neighbours. The merge logic is covered by unit tests.

None - feature is described above.

A screen recording will be added as a comment below.

None.

Signed-off-by: Åsmund Collin <aakjaergaard@gmail.com>

### DIFF
--- a/src/components/mixins/dashboard.ts
+++ b/src/components/mixins/dashboard.ts
@@ -18,13 +18,6 @@ import {
     mdiMulticast,
 } from '@mdi/js'
 import { afcIconLogo } from '@/plugins/afcIcons'
-import { GuiStateLayoutoption } from '@/store/gui/types'
-
-interface LayoutFixedPanel {
-    panel: GuiStateLayoutoption
-    predecessor: string | null
-    successor: string | null
-}
 
 @Component
 export default class DashboardMixin extends BaseMixin {
@@ -90,47 +83,5 @@ export default class DashboardMixin extends BaseMixin {
             default:
                 return mdiInformation
         }
-    }
-
-    mergeLayout(
-        storedPanels: GuiStateLayoutoption[],
-        oldPanels: GuiStateLayoutoption[],
-        newPanels: GuiStateLayoutoption[]
-    ): GuiStateLayoutoption[] {
-        const draggableNames = new Set(oldPanels.map((panel) => panel.name))
-        const storedByName = new Map(storedPanels.map((panel) => [panel.name, panel]))
-
-        const output = newPanels.map((panel) => storedByName.get(panel.name) ?? { name: panel.name, visible: true })
-        const movableNames = new Set(output.map((panel) => panel.name))
-
-        const fixedPanels: LayoutFixedPanel[] = []
-        let predecessor: string | null = null
-        storedPanels.forEach((panel) => {
-            if (draggableNames.has(panel.name)) {
-                fixedPanels.forEach((entry) => {
-                    if (entry.successor === null) entry.successor = panel.name
-                })
-
-                predecessor = panel.name
-                return
-            }
-
-            fixedPanels.push({ panel, predecessor, successor: null })
-        })
-
-        fixedPanels.forEach((entry) => {
-            let index = output.length
-
-            if (entry.predecessor !== null && movableNames.has(entry.predecessor)) {
-                index = output.findIndex((panel) => panel.name === entry.predecessor) + 1
-                while (index < output.length && !movableNames.has(output[index].name)) index++
-            } else if (entry.successor !== null && movableNames.has(entry.successor)) {
-                index = output.findIndex((panel) => panel.name === entry.successor)
-            }
-
-            output.splice(index, 0, entry.panel)
-        })
-
-        return output
     }
 }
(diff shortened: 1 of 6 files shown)


Hannott:
https://github.com/user-attachments/assets/6da08d06-5cfe-4805-9f85-8ca0e017c001


https://github.com/user-attachments/assets/8deba855-db66-4e4f-920e-868a4d202907

src/pages/Dashboard.vue
```diff
@@ -160,5 +131,79 @@ export default class PageDashboard extends Mixins(DashboardMixin) {
...
+        background-color 150ms ease-in-out;
+}
+
+.dashboard--dragging .dashboard-dropzone {
+    outline-color: var(--v-primary-base);
+    background-color: rgba(125, 125, 125, 0.06);
+}
+</style>
+
+<style>
```
meteyou: only use scoped style settings. you can use v-deep if you need to go deeper with the selector.

src/store/gui/getters.ts
```diff
@@ -158,6 +158,14 @@ export const getters: GetterTree<GuiState, RootState> = {
             return panels.filter((element) => allPossiblePanels.includes(element.name))
         },
 
+    // return the unfiltered layout, to save it back without losing hidden or unavailable panels
```
meteyou: the code should describe it by itself. so there should no comments needed. only for deeper description for magic number or something else, comments are "allowed".

src/pages/Dashboard.vue
```diff
@@ -121,36 +80,48 @@ import WebcamPanel from '@/components/panels/WebcamPanel.vue'
...
+        this.columns.forEach((column) => {
+            output[column.index] = this.$store.getters['gui/getPanels'](this.viewport, column.index, true)
+        })
+
+        return output
     }
 
-    get widescreenLayout3() {
-        return this.$store.getters['gui/getPanels']('widescreen', 3, true)
+    get dragGroup() {
```
meteyou: is this group really needed? there should be only 1 draggable viewport visible?

src/store/gui/getters.ts
Hannott: Agree, removed.

src/pages/Dashboard.vue
Hannott: Fixed, and noted for future reference.

src/pages/Dashboard.vue
Hannott: Yeah, my bad, I think that was something left over from an earlier iteration. Changed it to a constant.

Just the group prop itself has to stay, or else sortable wont drop into another list, and panels could only be reordered inside their own column.

src/components/mixins/dashboard.ts
```diff
@@ -84,4 +91,46 @@ export default class DashboardMixin extends BaseMixin {
                 return mdiInformation
         }
     }
+
+    mergeLayout(
```
meteyou: Pls move this funciton to the Dashboard.vue, because this function will only be used there. Mixins are for shared functions.

src/pages/Dashboard.vue
```diff
@@ -160,5 +128,77 @@ export default class PageDashboard extends Mixins(DashboardMixin) {
     extractPanelId(name: string) {
         return name.split('_')[1] ?? null
     }
+
+    dragStart(event: { item: HTMLElement }) {
+        this.resizeObserver = new ResizeObserver(() => {
```
meteyou, line 139: Can you pls explain why we need an ResizeObserver here? While a user move a panel, there should not be possible to resize the window? Or did i miss some facts/workflows here?

src/store/gui/getters.ts
```diff
@@ -158,6 +158,13 @@ export const getters: GetterTree<GuiState, RootState> = {
             return panels.filter((element) => allPossiblePanels.includes(element.name))
         },
 
+    getStoredPanels: (state) => (viewport: string, column: number) => {
```
meteyou: If this getter is only used in the Dashboard.vue, then i would move it to this file. In getters, there are normally only multi use getters.

meteyou (commented):
In this case, the location of the tests fits and I think we can merge these tests.

src/components/mixins/dashboard.ts
Hannott: Yeah, I'll move it.

src/store/gui/getters.ts
Hannott: Will move this too

src/pages/Dashboard.vue
Hannott: It's needed because of the different column widths. The real event.item gets resized when crossing column borders, and the fallback clone doesn't pick that up on its own, so the observer keeps it in sync.

PR #2613, opened 2026-07-27
fix(macros): stop marking existing macros as deleted in the settings

Hannott:
## Description

This PR fixes two bugs in the macros settings, plus a TypeError in `printer/getMacros`.

### 1. The search field labels group macros as deleted

`allMacros` in `SettingsMacrosTabExpert` returned the search-filtered macro list, and `existsMacro()` and `getMacroDescription()` used that same list to decide whether a macro still exists in the printer config.

Type anything into the Available Macros search box, and every macro in the edited group that does not match the search picks up the `Deleted macro` subtitle. Its color and visibility buttons vanish too, since `v-if="existsMacro(macro.name)"` turns false. The macros themselves are intact. The lookup was wrong.

Fix: `allMacros` now returns the unfiltered list. A new `filteredMacros` getter applies the search, and only `availableMacros` reads it.

### 2. Clearing the search field breaks both tabs

Both tabs pass `clearable` to their `v-text-field`. Vuetify resets the model to `null` when you click the clear icon, `searchMacros.toLowerCase()` throws, the computed fails, and the macro list stops rendering until you navigate away.

Fix: both tabs type `searchMacros` as `string | null` and coerce it before lowercasing.

### 3. TypeError in printer/getMacros when configfile is missing

`getMacros` defaulted `settings` to `null` when `state.configfile?.settings` was absent, then indexed it on the next line, so any state carrying `gcode_macro` objects without a populated configfile throws `Cannot read properties of null`. It now defaults to `{}`, the same shape as the fix in #2580 on this file.

I left the second `?? null` in `existsZtilt` alone. The `if (settings)` check right below it depends on the null.

No existing issue. I searched open and closed issues for `macro` and `deleted macro` and found nothing that matches.

I have no Moonraker instance to serve a build against, so I cannot supply before and after screenshots. Happy to add them if a maintainer wants them before merge.

Issue 1 reproduces in a few seconds:

1. Settings > Macros, set Management to Expert
2. Create a group and add two or more macros to it
3. Type part of one macro name into the Available Macros search box
4. On develop, every group macro that does not match shows `Deleted macro` and loses its color and visibility buttons. On this branch the group list stays put.

Unit tests cover all three fixes. Every regression test in them fails against develop and passes here:

- `tests/store/printer/getMacros.spec.ts` (8 tests)
- `tests/components/settings/settingsMacrosTabExpert.spec.ts` (10 tests)
- `tests/components/settings/settingsMacrosTabSimple.spec.ts` (7 tests)

`npx vitest run` reports 47 passing across 6 files. `npx vite build` and `eslint src` come back clean.

None.

### DIFF
--- a/src/components/settings/SettingsMacrosTabExpert.vue
+++ b/src/components/settings/SettingsMacrosTabExpert.vue
@@ -353,7 +353,7 @@ export default class SettingsMacrosTabExpert extends Mixins(BaseMixin, ThemeMixi
 
     private boolFormEdit = false
     private editGroupId: string | null = ''
-    private searchMacros: string = ''
+    private searchMacros: string | null = ''
 
     get groupColors() {
         return [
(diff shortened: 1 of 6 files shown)


meteyou:
Pls create 1 PR per issue/bug... I know its easy with a AI to fix all bugs at ones, but its worse to review.

Hannott:
Sure, will do!

PR #2617, opened 2026-07-31
fix(macros): prevent TypeError in printer/getMacros when configfile is missing

Hannott:
## Description

`getMacros` set `settings` to `null` when there was no configfile, then read `settings[propLower]` on the next line. So if a `gcode_macro` object arrives before the configfile is populated, it throws. Changed `?? null` to `?? {}`, same as #2580 did in this file.

Added tests for the getter. Two of them fail on develop.

No issue. Split out of #2613, which had three unrelated bugs in one PR. This one doesn't depend on the other two.

Nothing visual, it's a store getter.

No.

edit by mryel00:


meteyou:
do you really think 82 lines of testing code is a good solution for a 1 liner bugfix?

meteyou:
Pls also try to describe the PR text with your own words... i dont want to read a useless wall of text... i dont care if you have typos or bad english (my english is also very bad!). I just want to read human text and not a slop story...

Hannott:
Updated

Hannott:
> do you really think 82 lines of testing code is a good solution for a 1 liner bugfix?

Depends I guess. I never really concerned myself about LOC in tests, as long as they test something real. And, if they had been there, this bug would not have been, so yeah, maybe?

meteyou:
@Hannott did you read all tests and checked it yourself or just AI slop?

i need 10 times more time to read your PRs instead of fixing these bugs myself. I'm not here to just read AI slop. I will maintain this repo.

meteyou:
ok. i checked it now in detail again and I'm not against tests at all, but all stores have no tests right now (I know, we should add some), but your naming + location in this PR is "worst case". so pls remove them here. After the store refactoring (you can see all refactore(store) PRs and the feat/rework-init-process feature branch), i will think about a good file structure + naming for stores and will start to add some. Contributions are welcome then, but right now the tests in this PR are "wrong".

Hannott:
@meteyou Yeah, I do read all the code before commit, but having never added anything to this repo before, I guess it takes some back and fourth before I "get it right" in terms of matching the rest of the repo.

I'll remove the tests for now.

meteyou (approved):

PR #2618, opened 2026-07-31
fix(macros): keep the macro settings usable after clearing the search

Hannott:
## Description

Both macro settings tabs have `clearable` on the search field, and Vuetify sets the model to `null` when you hit the ✕. `searchMacros.toLowerCase()` then throws, the computed list dies, and the macro list stays empty until you close and reopen the settings.

Fix: type `searchMacros` as `string | null` in both tabs and fall back to `''` before lowercasing. Added `?? false` on the description half of the filter so it stays a boolean.

Repro on develop:

1. Settings > Macros (either mode)
2. Type something in the search box
3. Click the ✕
4. List is gone, console says `Cannot read properties of null (reading 'toLowerCase')`

Added tests for both tabs. The two null-search ones fail on develop.

No issue. Split out of #2613. Independent of #2617. The third one (`Deleted macro` label showing while searching) touches the same getter in the expert tab, so I'll open it on top of this branch.

No Moonraker instance here, so no before/after. Nothing visual changes anyway — it just stops the exception that emptied the list. Repro steps above.

No.

### DIFF
--- a/src/components/settings/SettingsMacrosTabExpert.vue
+++ b/src/components/settings/SettingsMacrosTabExpert.vue
@@ -353,7 +353,7 @@ export default class SettingsMacrosTabExpert extends Mixins(BaseMixin, ThemeMixi
 
     private boolFormEdit = false
     private editGroupId: string | null = ''
-    private searchMacros: string = ''
+    private searchMacros: string | null = ''
 
     get groupColors() {
         return [
@@ -398,11 +398,13 @@ export default class SettingsMacrosTabExpert extends Mixins(BaseMixin, ThemeMixi
     }
 
     get allMacros() {
+        const search = (this.searchMacros ?? '').toLowerCase()
         const macros = this.$store.getters['printer/getMacros'] ?? []
+
         return macros.filter((macro: PrinterStateMacro) => {
             return (
-                macro.name.toLowerCase().includes(this.searchMacros.toLowerCase()) ||
-                macro.description?.toLowerCase().includes(this.searchMacros.toLowerCase())
+                macro.name.toLowerCase().includes(search) ||
+                (macro.description?.toLowerCase().includes(search) ?? false)
             )
         })
     }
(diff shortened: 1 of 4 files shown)


src/components/settings/SettingsMacrosTabExpert.vue
```diff
@@ -398,11 +398,13 @@ export default class SettingsMacrosTabExpert extends Mixins(BaseMixin, ThemeMixi
...
     get allMacros() {
+        const search = (this.searchMacros ?? '').toLowerCase()
         const macros = this.$store.getters['printer/getMacros'] ?? []
+
         return macros.filter((macro: PrinterStateMacro) => {
             return (
-                macro.name.toLowerCase().includes(this.searchMacros.toLowerCase()) ||
-                macro.description?.toLowerCase().includes(this.searchMacros.toLowerCase())
+                macro.name.toLowerCase().includes(search) ||
+                (macro.description?.toLowerCase().includes(search) ?? false)
```
meteyou, line 407: why did you add here `?? false`?

src/components/settings/SettingsMacrosTabExpert.vue
```diff
@@ -353,7 +353,7 @@ export default class SettingsMacrosTabExpert extends Mixins(BaseMixin, ThemeMixi
 
     private boolFormEdit = false
     private editGroupId: string | null = ''
-    private searchMacros: string = ''
+    private searchMacros: string | null = ''
```
meteyou, line 356: Why did we need here `| null`? also the empty search field should be a string, or I'm wrong here?

meteyou:
Pls remove the tests. No settings page have tests right now. Also there are more store tests instead of setting tests in it.

src/components/settings/SettingsMacrosTabExpert.vue
Hannott: Just to make it boolean instead of boolean | undefined. Not important other than correctness imo.

src/components/settings/SettingsMacrosTabExpert.vue
Hannott: Vuetify's clearable prop resets the v-model to null, not blank. So when the clear icon is clicked, the computed list would crash.

But maybe a better solution is to use a backing field and get/set, so callers won't need to check.

PR #2623, opened 2026-08-01
fix(macros): keep the macro settings usable after clearing the search

Hannott:
Change searchMacros to use get and set, so that when nulling the searchfield, the string is set to empty, avoiding crash on the computed list

## Description

This PR fixes a defect where the macros list in the Expert view would be empty after clearing a search text


meteyou:
the null type was fine, I just didn't had this in mind. It was just a question from my side. It didn't mean that you have to implement "a overhead workaround" for that...

Hannott:
@meteyou I think this is better actually. Making it nullable serves no purpose other than having to check every call site. But it's up to you, just let me know which you prefer

meteyou (approved):

PR #2625, opened 2026-08-01
fix(macros): fix sorting function in macros expert mode

Hannott:
## Description

Dragging a macro more than one line up or down in list in settings, will just switch places with the item in that position. This PR fixes that issue, so an item will be placed where it is dropped. Now you dont have to swap every macro from the bottom up to get the last one on top.

https://github.com/user-attachments/assets/23b99977-ef3c-4afd-9a8e-4569f8ab6b1c

No


meteyou (approved):

meteyou:
Thx! It's funny that nobody reported this major bug until now!

PR #2627, opened 2026-08-04
refactor(store): rework printer/server initalisation

meteyou:
## Description

This PR is just to track the progress in the refactoring of the printer/server store and the init process of it.

none

none

none


Cstm3DBldr:
Not a bug you introduced, but the refactor makes it load-bearing:

`emitAndWait` promises are never settled when the socket closes. `waits` is only cleared by `removeWaitById`, which runs on a response or the optional timeout — `onclose`/`onerror`/`connect()` don't touch it. 41 of the 42 call sites pass no timeout.

On develop that barely matters, 5 call sites. Here it's 42, and `server/init` is a 13-deep await chain, so a Moonraker restart mid-init parks it forever: the await never returns, the catch never fires, and the dialog sits on whatever step it reached. `onOpen` then starts a second chain on reconnect.

Rejecting pending waits in `onclose` would cover it.

meteyou:
Thx for the hint, but I'm not finished this part of code, because I have to finish all child stores first.

When I'm finished this step, I will pay attention on the init process/workflow again.

(Sry if you get this message twice, I had network issues)

Issue #2238, opened 2025-07-05
2D g-code viewer like in Fluidd


Robbok3d:
### Requested feature:

The g-code viewer is slick and all but a g-code viewer/tracker like it is implemented in Fluidd is by far more useful. Is it possible to have a fluidd type 2d viewer as an option.

### Solves the following problem:

inconvenient g-code viewer without any analytical value or possibilities

### Additional information:

thanks to all developers!

meteyou:
Duplicate #870

Issue #870, opened 2022-06-06
Would like to see a option to put g-code viewer on the Dashboard just like webcam


karlalbury:
### Requested feature:

Add option to pull the Gcode Viewer Tracking window to the Dashboard.
find my self bouncing between the two a lot.

Include Render of what looking for.
![mainsail](https://user-images.githubusercontent.com/5194171/172127362-5afc1371-5cd4-4668-8c38-c458ea234ac5.jpg)

bnieves1994:
I would also love to see this.

piwi3910:
I would also like to see this, The feature exists in Fluidd, something similar in mainsail would be great

Sindarius:
I recommended against this because the 3D viewer is much more resource intensive compared to a 2d viewer. I think an alternative would be to bring some of the panels like progress and temps into the viewer page.

Nazgile94:
> I recommended against this because the 3D viewer is much more resource intensive compared to a 2d viewer. I think an alternative would be to bring some of the panels like progress and temps into the viewer page.

its optional, if you dont like it - dont add the widget to the dashboard

i +1 this, would be nice ! maybe with toggle for 3d/2d

PRGeno:
I agree with adding a 2D version of the GCode viewer to the dashboard especially for making Excluding Objects more useful and accessible.  I don't think the 3D viewer makes much sense on the dash, but for me a 2D viewer like Fluidd is a critical feature and one of the only reasons I typically use Fluidd over Mainsail.  

I think having the bed mesh on the dash like Fluidd would be nice to have too, but no where near as critical as a 2D GCode viewer.

Jeppedy:
I'd be happy with the OctoPrint-style GCode Viewer.  One layer at a time, showing what will be laid down for that layer, and allow scrubbing to see future/past layers.  A nice simple 2D viewer compared to what Mainsail has right now.

dot-Justin:
+1 this, would love to see.

sjoerd104:
+1 would be great, even octo's 2d one.

notwaffles:
+1

marrrkus:
+1

snuguru-maestro:
+1

PeterXQC:
+1

PR #2654, opened 2026-09-04
feat(dashboard): add live 2D G-code preview panel

RT530:
## Summary

- Adds a new "G-Code Preview" dashboard panel (and a matching dedicated page/sidebar entry, `mdiVideo2d`, next to the existing 3D G-Code Viewer's `mdiVideo3d`) that renders a live top-down 2D view of the toolpath for the file currently printing.
- G-code parsing (extracting X/Y extrusion moves into connected runs) runs in a Web Worker so a multi-megabyte file doesn't freeze the UI while it's being parsed.
- The printed vs. remaining portion of the path is split against `virtual_sdcard.file_position`, with a live pulsing marker at the current toolhead position (`motion_report.live_position`).
- Rendered as inline SVG scaled to the printer's actual bed dimensions (`toolhead.axis_minimum`/`axis_maximum`), including a non-square bed - no letterboxing. Includes a 25mm reference grid and red/green X=0/Y=0 axis lines.
- Also fixes a small pre-existing typo: the MMU panel's `Headline` locale string was `"Mmu"` while every other string in that section already said `"MMU"`.

## Test plan

- [x] `npm run lint` and `npm run build` both pass clean
- [x] Manually verified against a live printer (real dashboard, real bed dimensions, real print in progress) - printed/remaining split and live marker track correctly
- [x] Verified the panel behaves correctly with no file loaded, and once a print completes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0155khNsPnanEJu6Kd5ordUE

PR #2659, opened 2026-09-07
feat(dashboard): add live 2D G-Code Preview panel

RT530:
## Description

Adds a "G-Code Preview" panel to the dashboard: a live, top-down 2D view of the toolpath for the layer currently being printed, similar to Fluidd's 2D preview and with no new dependency.

- G-code parsing (grouping extrusion moves into per-Z-layer runs) runs in a Web Worker, off the main thread, so a multi-megabyte file doesn't freeze the UI while it is parsed. Layer detection ignores z-hop, so a lift-and-return travel isn't mistaken for a layer change.
- The panel follows whichever layer is currently printing (from `virtual_sdcard.file_position`) and draws only that layer, split into printed vs. remaining, with a marker at the live toolhead position (`motion_report.live_position`).
- Inline SVG scaled to the printer's real bed (`toolhead.axis_minimum`/`axis_maximum`), non-square beds included, with a 25 mm reference grid and red/green X=0/Y=0 axes.
- Two options: **Print preview** (on by default; off draws only what has actually been printed) and **Show move path** (off by default; overlays non-extruding travel moves as a dashed line trailing the toolhead).
- Also fixes a small pre-existing typo: the MMU panel's `Headline` locale string read `"Mmu"` while every other string in that section already said `"MMU"`.

Per review feedback this is **only** a dashboard panel - the separate page, its route and its sidebar entry have been dropped, along with the page-only sizing code that came with them.

Implements the dashboard 2D preview requested in #870 (open since 2022). Also related to its duplicate, #2238.

Tested against a real sliced file (6.4 MB, 78 layers - the parser's layer count matches the slicer's own `layer_count` metadata exactly). Shown mid-print on layer 30 with both options enabled: printed path in blue, remaining in grey, travel moves dashed, toolhead marked in orange.

G-Code Preview dashboard panel

The panel is an ordinary dashboard panel, so it follows the dashboard's own responsive layout on smaller screens; happy to add a mobile capture as well if that would help.

None.

Signed-off-by: Ricky Tsai <ricky@rtnztech.com>


meteyou:
was this PR intended, or have you just issues to control your clanker?

RT530:
@meteyou thanks for the feedback on #2654! I've rebased and updated this PR (#2659, since #2654 had fallen behind `develop`):

- Description now follows the PR template (Description / Related Tickets / Screenshots / post-deployment), with desktop + mobile screenshots included.
- Linked the long-standing feature request this implements (#870).
- Picked up two improvements since #2654: the preview is now **per-layer** (it shows only the layer currently being printed rather than flattening the whole file, and layer detection ignores z-hop so a lift-and-return travel can't be mistaken for a layer change), plus an optional **"Show move path"** toggle for travel moves, off by default.

One heads-up so the diff doesn't confuse you: because this PR is closed, GitHub has its head pinned at `c0f667b0`, so the "Files changed" tab still shows only the original three commits - the two improvements above are on the branch (tip `514e6405`) but won't appear in the diff until the PR is reopened, at which point it re-syncs to the branch automatically.

No rush at all - happy to make any other changes, and I understand the vouch requirement.

meteyou:
ok, you dont have to create new PRs, just add your commits to the branch and the PR will be updated.

first of all, i would not add another page and just would add an panel for the dashboard (like the description in the FR). Did you test already more complex gcode files with that?

meteyou:
vouch @RT530

RT530:
Thanks for the vouch @meteyou, and for the direction.

**Dropped the extra page.** It's a dashboard panel only now - the page, its route and its sidebar entry are gone, along with the page-only sizing code that came with them, so there's no dead code left behind. Description updated to match.

**On more complex files:** yes. Tested against a real sliced file (6.4 MB, 78 layers). Two things worth reporting:

- The parser's layer count matches the slicer's own `layer_count` metadata exactly (78).
- Parsing it takes ~150 ms in the worker, and drawing one layer is ~2 ms. The thing that actually hurt was Vue deep-observing ~175k parsed point objects, so the parsed result is now frozen before being stored - that's what makes a file this size comfortable rather than the rendering itself.

**CodeRabbit's six findings are all fixed** - they were good catches, including two I'd call real bugs:

- extruding `G2`/`G3` were classified as travel, so an arc-welded file (common with Klipper) would have recorded its whole toolpath as travel and dropped the layer entirely
- lead-in travel was attributed to the layer being left rather than the one being entered
- offsets counted UTF-16 code units rather than UTF-8 bytes, which would drift against `virtual_sdcard.file_position` on any file with a non-ASCII comment
- the chart's throttled progress never picked up its initial value, so a print that was idle or finished drew the whole path as not-yet-printed
- a superseded load could overwrite a newer one (cancelling can't recall an already-fulfilled request), now guarded with a load id
- the size limit was enforced after buffering the whole body; it now checks `content-length` first

I verified the parser fixes against both the real file and a small non-ASCII/arc fixture rather than by inspection.

Happy to keep adjusting - and if you'd prefer the two checkboxes moved, renamed or dropped, just say.

meteyou:
ok. i will review/test this PR in the next days, then i can give more/deeper feedback

